PR #21, opened 2018-01-23
noticket - logger skeleton implemented

ratoshniuk:


lib/logger/build.sbt
```diff
@@ -0,0 +1,7 @@
+libraryDependencies ++=
+  Seq(
+//      "com.github.pureconfig" %% "pureconfig" % "0.9.0"
+//    , "com.github.pureconfig" %% "pureconfig-enum" % "0.9.0"
+    "org.slf4j" % "slf4j-api" % "1.7.25"
```
pshirshov: slf4j support should be implemented in a separate module

lib/logger/build.sbt
```diff
@@ -0,0 +1,7 @@
+libraryDependencies ++=
+  Seq(
+//      "com.github.pureconfig" %% "pureconfig" % "0.9.0"
```
pshirshov: we are not going to use pureconfig, it has all the flaws of original typesafe-config

lib/logger/src/main/scala/org.bitbucket.pshirshov.izumi/config/ConfigLoader.scala
```diff
@@ -0,0 +1,9 @@
+package org.bitbucket.pshirshov.izumi.config
+
+import org.bitbucket.pshirshov.izumi.logger.Log
+import org.bitbucket.pshirshov.izumi.logger.Models.{LoggingConfig, LoggingRule}
+
+trait ConfigLoader {
+
+  lazy val loggingConfig: LoggingConfig = LoggingConfig(rules = Seq(LoggingRule("", Set(Log.Level.Debug))))
```
pshirshov: You shouldn't think about logger configs at this stage, you should concentrate on API. Also `lazy val` is not a good thing.

lib/logger/src/main/scala/org.bitbucket.pshirshov.izumi/logger/LogSink.scala
```diff
@@ -0,0 +1,38 @@
...
+import scala.annotation.tailrec
+
+trait LogSink {
+
+  def messageParser(msg: Log.Message): Map[String, Any] = {
+    def truncateKey(s: String) = {
+      "(\\w+)=".r.findFirstIn(s).map(_.dropRight(1))
+    }
+    @tailrec
+    def parse(keysIterator: Iterator[String]
```
pshirshov: You may avoid recursion here. It may be a premature optimization for now though it's easy to parse everything during a single pass.

lib/logger/src/main/scala/org.bitbucket.pshirshov.izumi/logger/LogSink.scala
```diff
@@ -0,0 +1,38 @@
+package org.bitbucket.pshirshov.izumi.logger
+
+import scala.annotation.tailrec
+
+trait LogSink {
+
+  def messageParser(msg: Log.Message): Map[String, Any] = {
```
pshirshov: Parsing should happen BEFORE sink.

Logger message should have the following signature: `(args: Map[String, Any], template: String, message: String)` where message is a rendered message, template is original message with placeholders instead of arguments. Sink should accept these messages.

Your parsing should happen withing your Logger implementation.

pshirshov (changes requested):

pshirshov:
Closing, all the code adopted in `feature/logs`

PR #86, opened 2018-03-02
add base classes for http4s transport impl

dnaumenko:


pshirshov (changes requested):
You need to move all the transport implementation here

PR #222, opened 2018-06-19
Feature/support more type shapes in TagK

neko-kai:
Also add path-dependent types tests (currently macros work for 2/3 cases, runtime reflection doesn't work for any)


pshirshov:
To be fixed:

```
[error] /home/travis/build/pshirshov/izumi-r2/distage/distage-model/src/main/scala/com/github/pshirshov/izumi/distage/model/reflection/universe/WithDITypeTags.scala:165:22: type mismatch;
[error]  found   : WithDITypeTags.this.Tag[T]
[error]  required: WithDITypeTags.this.u.WeakTypeTag[_]
[error]       Tag.appliedTag(tag, List(tag.tag))
[error]                      ^
[info] WithDITypeTags.this.Tag[T] <: WithDITypeTags.this.u.WeakTypeTag[_]?
```

PR #368, opened 2018-08-31
use daemon thread for queueing sink

dnaumenko:


pshirshov:
Already fixed better way.

Issue #320, opened 2018-07-31
Human-friendly plan printer


pshirshov:
At the moment we have a lot of noise in our plan representations. 

Most annoying things are package names.

So, we should implement a stateful `PlanRenderer` which would use a smart heuristic to make plans readable:

1. When we print a classname and it is unique across plan keys we should use just class name
2. We should print FQN otherwise

pshirshov:
Still has to be improved. Also we should drop `Formattable` crap. Though nice enough to go with.

PR #406, opened 2018-10-02
add formatter for human-readable plan rendering

dnaumenko:
Fix for #320


distage/distage-core/src/test/scala/com/github/pshirshov/izumi/distage/injector/PlanTest.scala
```diff
@@ -0,0 +1,21 @@
...
+class PlanTest extends WordSpec with MkInjector {
+  "OrderedPlan has an human-readable toString impl" in {
+    val injector = mkInjector()
+    val plan = injector.plan(new ModuleDef {
+      many[JustTrait]
+        .add(new Impl1)
+    })
+
+    val formatted = new PlanFormatter().format(plan)
+    println(formatted)
```
pshirshov: `println`

PR #455, opened 2018-11-13
feature/noticket-memoization-fix

Caparow:


neko-kai:
👍

distage/distage-roles/src/main/scala/com/github/pshirshov/izumi/distage/roles/launcher/RoleStarterImpl.scala
```diff
@@ -21,16 +21,22 @@ class RoleStarterImpl
   executors: Set[ExecutorService],
   integrations: Set[IntegrationComponent],
   logger: IzLogger,
+  resourceCollection: ResourceCollection = new NullResourceCollection
```
neko-kai: Creating your own dependencies is the problem DI is supposed to solve in the first place, please don't do that...

distage/distage-roles-api/src/main/scala/com/github/pshirshov/izumi/distage/roles/roles/ResourceCollection.scala
```diff
@@ -0,0 +1,22 @@
...
+
+  def getComponents: Set[RoleComponent]
+
+  def getExecutors: Set[ExecutorService]
+
+  def processContext(context: Locator): Unit
+
+  def transformPlanElement(op: ExecutableOp): ExecutableOp
+
+  def close(closeable: AutoCloseable): Unit
```
neko-kai: This method is never used, and also `getX` methods expose internals in an extreme way making it *even* harder to refactor in the future than currently.

I propose to change the interface to just

```scala
trait ResourceCollection {
  def canStart(resource: Any): Boolean

  def registerStarted(resource: Any): Unit

  def canClose(resource: Any): Boolean

  def processContext(context: Locator): Unit

  def transformPlanElement(op: ExecutableOp): ExecutableOp
}
```

This should drastically simplify logic - `ResourceCollection` wouldn't need to duplicate any of the lifecycle functions already done by `RoleStarter`, `RoleStarter` would just need to query `ResourceCollection` before starting/stopping any AutoCloseables/RoleComponents/ExecutorServices and call `registerStarted` after starts - where `registerStarted` will do nothing for unmemoized instances, but start returning `canStart=false` for memoized instances.

neko-kai (changes requested):

distage/distage-testkit/src/main/scala/com/github/pshirshov/izumi/distage/testkit/DistageTests.scala
```diff
@@ -38,7 +39,7 @@ trait DistageTests {
     di(providerMagnet)
   }
 
-  protected def di(f: ProviderMagnet[Any]): Unit = {
+  protected def di(f: ProviderMagnet[Any]): Unit = synchronized {
```
neko-kai: Is this necessary?

distage/distage-testkit/src/main/scala/com/github/pshirshov/izumi/distage/testkit/TestComponentsLifecycleManager.scala
```diff
@@ -0,0 +1,78 @@
...
+    component.start()
+    deque.pop().discard()
+    deque.push(ComponentLifecycle.Started(component))
+  }
+
+  override def startComponents(): Unit = {
+    components.foreach { component =>
+      component.synchronized {
+        if (resourceCollection.isMemoized(component)) {
+          if (!isMemoizedComponentStarted(component)) {
```
neko-kai: I'd rather put start info into `resourceCollection` too, so that all dirty stuff is in one place and we don't have to keep two copypastas of `ComponentLifecycleManager` logic

distage/distage-testkit/src/main/scala/com/github/pshirshov/izumi/distage/testkit/DistageTests.scala
aliubymov-playq: Just forgot to remove

distage/distage-testkit/src/main/scala/com/github/pshirshov/izumi/distage/testkit/TestComponentsLifecycleManager.scala
aliubymov-playq: @kaishh we need to rewrite this `ComponentLifecycleManager` due to synchronization over components, other `ComponentLifecycle` collection.

neko-kai: It's okay to synchronize and use ConcurrentLinkedDeque in main code. I think it would be good to just replace old impl with this and move `isMemoizedComponentStarted` logic into `ResourceCollection`

aliubymov-playq: I think that sharing resource collections with main role starter and binding it with DI is not a good idea. So I've just replaced an old collection with a new one and added a synchronization to be able to override `startComponents()` using memoizing components.

neko-kai: Don't share, just add methods to `ResourceCollection` like this:

```scala
trait ResourceCollection {
  def register(started: ComponentLifecycle.Started): Unit =
    if (isMemoized(started.component) && !deque.contains(...))
      deque.put

  def isAlreadyStarted(component: Any): Unit = 
      deque.contains(...)
}
```

And keep an independent deque inside ResourceCollection

neko-kai: Hmm, okay I guess, maybe it's not a good idea to use ResourceCollection in main starter (it's supposed to be NullResourceCollection there though)

neko-kai (approved):

distage/distage-testkit/src/main/scala/com/github/pshirshov/izumi/distage/testkit/TestComponentsLifecycleManager.scala
```diff
@@ -0,0 +1,39 @@
+package com.github.pshirshov.izumi.distage.testkit
+
+import com.github.pshirshov.izumi.distage.roles.launcher.{ComponentLifecycle, ComponentsLifecycleManagerImpl}
+import com.github.pshirshov.izumi.distage.roles.roles.RoleComponent
+import com.github.pshirshov.izumi.distage.testkit.MemoizingDistageResourceCollection.memoizedComponentsLifecycle
```
neko-kai: You probably still want to add a method in ResourceCollection instead of directly accessing mutable globals. You might want to make everything in the `MemoizingDistageResourceCollection` object private.

neko-kai:
LGTM

PR #497, opened 2019-03-14
Add BIO(_) syntax sugar for BIO.syncThrowable 

neko-kai:
Add BIO postfix syntax .attempt, .fromEither, .fromOption


pshirshov (approved):

fundamentals/fundamentals-bio/src/main/scala/com/github/pshirshov/izumi/functional/bio/BIO.scala
```diff
@@ -46,11 +46,15 @@ trait BIO[R[+ _, + _]] extends BIOInvariant[R] {
   @inline override def leftFlatMap[E, A, E2](r: R[E, A])(f: E => R[Nothing, E2]): R[E2, A]
 
   @inline override def flip[E, A](r: R[E, A]) : R[A, E]
+
+  @inline final def apply[A](effect: => A): R[Throwable, A] = syncThrowable(effect)
```
ratoshniuk, line 52: is it obvious to make apply as alias of syncThrowable? zio has apply for sync, cats has for delay.

neko-kai: ZIO apply is for syncThrowable - https://github.com/scalaz/scalaz-zio/blob/master/core/shared/src/main/scala/scalaz/zio/ZIO.scala#L1580
delay is also syncThrowable

PR #498, opened 2019-03-14
WIP Feature/resource

neko-kai:
Support monadic effect and resource bindings, e.g.:

```scala
// effect bindings
make[Ref[F, A]].named("global-variable").fromEffect(Ref[F].of(a))

// resource bindings
make[org.http4s.Client[F]].fromResource(ec => BlazeClientBuilder[F](ec).resource)
make[DynamoClient[F]].fromResource[DynamoResource[F]]

class DynamoResource[F[_]: Sync](tables: Set[DynamoTable])
  extends DIResource[F, DynamoClient[F]] {

  override val allocate = for {
    client <- newClient[F]
    _ <- tables.toList.traverse(table => client.execute(table.ddl))
  } yield client

  override def deallocate(client: DynamoClient[F]): F[Unit] = client.close
}
```

This feature obviates `RoleComponent`, `RoleStarter`, `AutoCloseable`/`ExecutorService` support etc. and makes lifecycle a core concept instead of an add-on


distage/distage-core/src/main/scala/com/github/pshirshov/izumi/distage/planning/extensions/GraphDumpObserver.scala
```diff
@@ -149,27 +149,35 @@ final class GraphDumpObserver
...
                     "ref"
                 }
-              case RuntimeDIUniverse.Wiring.FactoryMethod(_, _, _) =>
-                "fdef"
 
+              case RuntimeDIUniverse.Wiring.Factory(_, _, _) =>
+                "factory"
               case RuntimeDIUniverse.Wiring.FactoryFunction(_, _, _) =>
-                "ffun"
+                "factoryfun"
```
ratoshniuk, line 170: camelcase maybe?

PR #524, opened 2019-04-29
 Fix #500 Split BIO into trait hierarchy, add conversions to cats-effect hierarchy

### LINKED ISSUE #500: Split BIO into traits
@kaishh , do we always need to tag BIO as an Interface of whole methods? 

Maybe we could split it up as BIOMonad, BIOFunctor, BIOEffect, etc

neko-kai:


neko-kai:
* remove BIOInvariant, BIOAsyncInvariant
* add `<*`, `.peek`, `map2`, `guarantee`

Issue #536, opened 2019-05-29
IzManifest return wrong version of an artifact.


VladPodilnyk:
Here is a way to reproduce a bug:
```scala
object A extends App {
  import scala.collection.JavaConverters._
  implicit val clock : Clock[Identity] = new Clock.Standard
 
  val a = IzManifest.manifest[Version.type]().map(_.getMainAttributes.asScala)
  val b = IzManifest.manifest[Clock.type]().map(_.getMainAttributes.asScala)
  val c = IzManifest.manifest[cats.effect.IO.type]().map(_.getMainAttributes.asScala)

  println(a)
  println(b)
  println(c)
}
```
Here is an output of the snippet above:
```
Some(Map(Manifest-Version -> 1.0, Specification-Title -> appstatus-api, Specification-Version -> 0.49.0-SNAPSHOT, Specification-Vendor -> net.playq.api, Implementation-Title -> appstatus-api, Implementation-Version -> 0.49.0-SNAPSHOT, Implementation-Vendor -> net.playq.api, Implementation-Vendor-Id -> net.playq.api, Implementation-URL -> https://playq.net))

Some(Map(Manifest-Version -> 1.0, Implementation-Title -> fundamentals-platform, Implementation-Version -> 0.7.0, Specification-Vendor -> com.github.pshirshov.izumi.r2, X-Git-Head-Rev -> fdb233d17297fc4de2149b45de36c7c05e214fc3, Specification-Title -> fundamentals-platform, Implementation-Vendor-Id -> com.github.pshirshov.izumi.r2, Implementation-Vendor -> com.github.pshirshov.izumi.r2, X-Built-By -> liyon, X-Git-Branch -> fdb233d17297fc4de2149b45de36c7c05e214fc3, X-Build-JDK -> 1.8.0_181, X-Build-Timestamp -> 2019-03-31T20:11:41.143Z[Etc/UTC], X-Build-SBT -> 1.2.8, X-Build-Scala -> 2.12.8, X-Git-Repo-Is-Clean -> false, X-Version -> 0.7.0, Specification-Version -> 0.7.0, Implementation-URL -> https://izumi.7mind.io))

[...]

PR #538, opened 2019-05-30
Fix IzManifest version retrieving from an artifact 

VladPodilnyk:
#536 
Some artifacts does not have 'X-Version' attribute,
so as a fallback option we search for 'Implementation-Version'.


VladPodilnyk:
@ratoshniuk, could your review this when you will have a free time?

PR #558, opened 2019-06-25
WIP try to catch IntegrationCheckException from inside F

neko-kai:
to allow DIEffectRunner/BIORunner to return a decorated exception instead of original

not finished (have to expose `IO.halt` to recover original trace after `.sandbox` ._.)


neko-kai:
obsoleted by new distage-testkit

Issue #520, opened 2019-04-23
Delete runtime reflective config reader. Use circe-config or pureconfig


neko-kai:
Seems like runtime-reflection based config reader impacts test start performance negatively, we may replace it with a compile-time reader with no loss of functionality

neko-kai:
Due to unavoidable concurrency bugs and the general EOL of `scala-reflect` (https://github.com/scala/bug/issues/10766) we should limit its usage. We're hitting concurrency issue on large config case classes, so runtime config reader has to go. In this case, replacing magic runtime codec with first-class typeclass-based codecs has no real downsides, especially if we can use a fast deriver (e.g. `circe-derivation` for `circe-config`)
@Caparow @pshirshov

PR #582, opened 2019-07-17
Scala 2.13 WIP

neko-kai:
compiles, but tests fail


neko-kai:
Unfortunately we're hitting another scala-reflect bug scala/bug#11645 may make distage-config less reliable in sealed traits (although https://github.com/7mind/izumi/issues/520)...

PR #587, opened 2019-07-18
Zero-import BIO syntax via implicit punning

neko-kai:


idealingua-v1/idealingua-v1-runtime-rpc-http4s/src/test/scala/com/github/pshirshov/izumi/idealingua/runtime/rpc/http4s/fixtures/DummyAuthorizingDispatcher.scala
```diff
@@ -1,7 +1,7 @@
 package com.github.pshirshov.izumi.idealingua.runtime.rpc.http4s.fixtures
 
 import com.github.pshirshov.izumi.functional.bio.BIO
-import com.github.pshirshov.izumi.functional.bio.BIO._
+import com.github.pshirshov.izumi.functional.bio.BIO
```
dnaumenko: duplicate

distage/distage-model/src/main/scala/com/github/pshirshov/izumi/distage/model/monadic/DIEffect.scala
```diff
@@ -2,7 +2,7 @@ package com.github.pshirshov.izumi.distage.model.monadic
 
 import cats.effect.ExitCase
 import com.github.pshirshov.izumi.distage.model.monadic.FromCats._Sync
-import com.github.pshirshov.izumi.functional.bio.{BIO, BIOExit}
+import com.github.pshirshov.izumi.functional.bio.{BIO, BIOExit, F}
```
dnaumenko: Quite hard to notice it here, maybe move to DIEffect method. It's named as type param by purpose as I understand, but not sure if it's a good idea to mix different concepts.

fundamentals/fundamentals-bio/src/main/scala/com/github/pshirshov/izumi/functional/bio/BIO.scala
```diff
@@ -156,7 +155,7 @@ object BIOPanic {
   @inline final def apply[F[+_, +_]: BIOPanic]: BIOPanic[F] = implicitly
 }
 
-trait BIO[F[+_, +_]] extends BIOPanic[F] {
+private[bio] trait BIO0[F[+_, +_]] extends BIOPanic[F] {
```
dnaumenko: Can't say I follow the trick here :/

fundamentals/fundamentals-bio/src/main/scala/com/github/pshirshov/izumi/functional/bio/package.scala
```diff
@@ -1,12 +1,157 @@
...
 
-import com.github.pshirshov.izumi.functional.mono.SyncSafe
+  import com.github.pshirshov.izumi.functional.mono.SyncSafe
 
-package object bio {
-  type SyncSafe2[F[_, _]] = SyncSafe[F[Nothing, ?]]
 
-  object SyncSafe2 {
-    def apply[F[_, _] : SyncSafe2]: SyncSafe2[F] = implicitly
+  package object bio extends B1 {
```
dnaumenko: Can't say I follow the trick here :/

dnaumenko: As I understand, the idea is to move implicits conversions for BIO to package object since, but I miss the point of why we have to use type aliases and lazy vals. What's reason behind the SyncSafe2, etc.

Another concern, since package object will be deprecated in Scala 3.0, will this approach be backward compatible or not?

dnaumenko (approved):

distage/distage-model/src/main/scala/com/github/pshirshov/izumi/distage/model/monadic/DIEffect.scala
neko-kai: moved directly to implicit def as paramter `(implicit F: BIO[F])`

neko-kai:
@dnaumenko 
Thanks for the feedback! The point of the trick is to remove the annoying `BIO._` import everywhere, by punning syntax conversions to the same as the type, they can always follow the type around. example: https://scastie.scala-lang.org/EAdQeqJxTxCdmSlOvBmcag

This allows Intellij auto-import to include all relevant syntax without having to write manual imports.
The same reason applies to moving `def F` from a simple dependent method to a bunch of self-conversions -> Intellij IDEA can find and auto-import only objects, but not defs or vals.

PR #649, opened 2019-09-06
Try to update to circe-derivation-0.12.0-M7

neko-kai:


neko-kai:
Test failures - https://github.com/circe/circe-derivation/issues/103

PR #743, opened 2019-11-22
Fix japanese typo in docs + README

cronokirby:
The female name 泉水 was used instead of the noun 泉 which means "spring" (both are pronounced "izumi").
See https://jisho.org/search/%E6%B3%89%E6%B0%B4 for more info.

(I also replaced "jap." with "jp." as an abbreviation of "japanese", "jap" is considered a racial slur)

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 What is it?
 ===========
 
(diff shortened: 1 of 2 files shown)


neko-kai:
@cronokirby 
Thanks for your contribution! However, I've reverted the change to kanji, since the use of a given name is intentional ;)

PR #770, opened 2019-12-05
Compute hashcode only based on short type name

neko-kai:
Because after combining higher-kinded LTTs their lambda parameter names mismatch:

`TagK[Either[Throwable, ?]]` = `%0 => Either[Throwable, 0]`
but `[F[_, _]: TagKK] => TagK[F[Throwable, ?]]` = `%1 => Either[Throwable, 1]`


neko-kai:
☝️ @pshirshov

PR #807, opened 2020-01-02
Feature/weak reference memoization

Caparow:


distage/distage-testkit-scalatest/src/test/scala/izumi/distage/testkit/distagesuite/DistageTestExample.scala
```diff
@@ -37,15 +39,60 @@ class DistageTestExampleBIO extends DistageBIOSpecScalatest[zio.IO] with Distage
...
+
+    "support named weak sets" in {
+      (
+        set: Set[SetExample] @Id("set-id"),
+        s1: SetElement1,
+        s2: SetElement2,
+        s3: SetElement3,
+      )=>
+        Quirks.discard(s1, s2, s3)
+        F.maybeSuspend(assert(set.size == 3))
```
neko-kai, line 105: Could you add a test that summons Just Set, without set elements, and checks that size = 0?

distage/distage-testkit-core/src/main/scala/izumi/distage/testkit/services/dstest/DistageTestRunner.scala
```diff
@@ -73,10 +74,17 @@ class DistageTestRunner[F[_]: TagK]
         // here we find all the shared components in each of our individual tests
         val sharedKeys = testPlans.map(_._2).flatMap {
           plan =>
-            plan.steps.filter(env memoizationRoots _.target).map(_.target)
+            val memoized = plan.steps.filter(env memoizationRoots _.target).map(_.target)
+            val weak = plan.steps.collect {
+              // here we adding each weak reference into shared keys (we assume that if we have a weak reference in one test plan - it's should be bind)
+              case w: WiringOp.ReferenceKey if w.wiring.weak => w.target
```
neko-kai: This makes ALL weak references into roots of the memoization plan - this is too broad and will result in a huge chunk of the graph being memoized.

distage/distage-testkit-core/src/main/scala/izumi/distage/testkit/services/dstest/DistageTestRunner.scala
```diff
@@ -73,10 +74,17 @@ class DistageTestRunner[F[_]: TagK]
...
+              // adding sets into shared keys if one of the members is preserving in test plan (we should build set with a weak reference before tests start)
+              case c: CreateSet if c.members.exists(plan.index.keys.toList.contains) => c.target
```
neko-kai: I think you can't solve the problem by adding roots, because roots are _new components_ that result in new large subgraphs being added to shared locator. If you add all sets with weak members into roots, the result graph will be much-much larger than the original graph of "only dependencies of memoizationRoots"

I think what needs to be done is pass a custom weakness predicate to `TracingDIGC`,
such that **during GC** when it encounters a weak key it may be considered referenced, when it's present in test plans. This would make it include ONLY those weak references that are in sets that ARE dependencies of memoizationRoots, not ALL sets.

distage/distage-testkit-scalatest/src/test/scala/izumi/distage/testkit/distagesuite/DistageTestExample.scala
Caparow: Yup, sure.

distage/distage-testkit-core/src/main/scala/izumi/distage/testkit/services/dstest/DistageTestRunner.scala
Caparow: Hm, okay then, I'll take a look into.
I've thought that it would be okay to store in memoizationRoots whole set of weak references.

distage/distage-testkit-scalatest/src/test/scala/izumi/distage/testkit/distagesuite/DistageTestExample.scala
neko-kai: You may want tests for memoized sets + non-memoized sets and assert that their behavior is different wrt inclusion on weak references that are only summoned in *other* tests

distage/distage-testkit-core/src/main/scala/izumi/distage/testkit/services/dstest/DistageTestRunner.scala
neko-kai: It's 'okay', but it won't behave as expected. e.g. when you have memoizationRoots=empty your weak DDLs will still be memoized, and with them everything that depends on them - i.e. a huge chunk of the graph. You'd expect that `memoizationRoots=Set.empty` means no memoization at all, but this is not the case when additional roots are added synthetically...

Caparow: @neko-kai may you check new implementation ?

distage/distage-testkit-core/src/main/scala/izumi/distage/testkit/services/dstest/DistageTestRunner.scala
```diff
@@ -170,7 +160,8 @@ class DistageTestRunner[F[_] : TagK]
                   val allSharedKeys = mainSharedLocator.allInstances.map(_.key).toSet
 
                   val integrations = testplan.collectChildren[IntegrationCheck].map(_.target).toSet -- allSharedKeys
-                  val newtestplan = testInjector.trisectByRoots(appmodule.drop(allSharedKeys), testplan.keys -- allSharedKeys, integrations)
+                  val predicate: WeaknessPredicate = testplan.steps.map(_.target).contains
+                  val newtestplan = testInjector.trisectByRoots(appmodule.drop(allSharedKeys), testplan.keys -- allSharedKeys, integrations, predicate)
```
neko-kai: The predicate should be applied in `shared`. Here it's no-op (non-weak keys that are referenced in test plan are already referenced in test plan, even without the predicate)

pshirshov (changes requested):
The whole idea is wrong. This can be done without modifications in planner input, this can be done purely at plan trisplit level.

distage/distage-testkit-core/src/main/scala/izumi/distage/testkit/services/dstest/DistageTestRunner.scala
Caparow: Yeah, I found error in my tests.

distage/distage-core-api/src/main/scala/izumi/distage/model/definition/ModuleBase.scala
```diff
@@ -59,6 +59,15 @@ object ModuleBase {
       T.make(module.bindings.map(f))
     }
 
+    def foldWith[B](b: B)(f: (Binding, B) => (Binding, B)): (T, B) = {
```
Caparow: I've add here new method, to log strengthen bindings without using mutable collections.

distage/distage-framework/src/main/scala/izumi/distage/framework/services/ConfigLoader.scala
```diff
@@ -29,10 +29,11 @@ object ConfigLoader {
...
     */
   class LocalFSImpl(
-    logger: IzLogger,
-    primaryConfig: Option[File],
-    roleConfigs: Map[String, Option[File]],
-  ) extends ConfigLoader {
+                     logger: IzLogger,
+                     primaryConfig: Option[File],
+                     roleConfigs: Map[String, Option[File]],
+                     configOverride: AppConfig => AppConfig = identity
```
Caparow: @neko-kai @pshirshov 
Is this looks good? I don't wanna add new method here, because it's redundant.

distage/distage-testkit-scalatest/src/test/scala/izumi/distage/testkit/distagesuite/DistageTestExample.scala
```diff
@@ -37,15 +40,66 @@ class DistageTestExampleBIO extends DistageBIOSpecScalatest[zio.IO] with Distage
...
+        s1: SetElement1,
+        s2: SetElement2,
+        s3: SetElement3,
+      ) =>
+        Quirks.discard(s1, s2, s3)
+        F.maybeSuspend(assert(set.size == 3))
+    }
+
+    "return memoized weak set with have whole list of members even if test does not depends on them" in {
+      set: Set[SetElement] =>
```
Caparow: @neko-kai does we have any chance to skip this test if it runs without other? Without registry singleton using ?

distage/distage-framework/src/main/scala/izumi/distage/framework/services/ConfigLoader.scala
neko-kai: @Caparow This parameter is redundant given `ConfigLoader.map` method here - https://github.com/7mind/izumi/pull/807/files#diff-5c8b9d105825c5921af472f1d4cce8b4R17

neko-kai: @Caparow Also, could you update the comment with default config locations here - https://github.com/7mind/izumi/pull/807/files#diff-5c8b9d105825c5921af472f1d4cce8b4R25 ?

distage/distage-testkit-scalatest/src/test/scala/izumi/distage/testkit/distagesuite/DistageTestExample.scala
neko-kai: @Caparow 
You could use `assume` instead of `assert`, then the test would skip instead of failing when the predicate fails.

But, yeah, the test will fail if it's run individually, instead of with all tests:

```scala
Test failed
ScalaTestFailureLocation: izumi.distage.testkit.distagesuite.DistageTestExampleBase at (DistageTestExample.scala:100)
org.scalatest.exceptions.TestFailedException: Set() had size 0 instead of expected size 3
	at org.scalatest.Assertions.newAssertionFailedException(Assertions.scala:530)
```

I think instead of skipping this test, we may probably want to always plan using _all_ tests in a test env and only filter them out later, after we've strengthened the weak components...

/cc @pshirshov

distage/distage-testkit-core/src/main/scala/izumi/distage/testkit/services/dstest/SpecEnvironment.scala
```diff
@@ -20,6 +20,7 @@ trait SpecEnvironment {
 
   def contextOptions: PlanningOptions
 
+  def configOverrides: AppConfig => AppConfig
```
neko-kai, line 23: IMHO this is also unnecessary, because of `override` + `ConfigLoader#map`

distage/distage-core-api/src/main/scala/izumi/distage/model/definition/ModuleBase.scala
neko-kai: Might as well add a more general `foldLeft[B](f: (B, Binding) => B): B` and base this on it. Also I guess you need to specify the direction of the accumulator parameter, because

1. A Set fold may not actually be unordered, because there's `ListSet` & `LinkedHashSet` - ordered Sets in stdlib
2. There are still two variants of where to place the accumulator parameter in the stdlib. – currently the accumulator parameter is to the right, as in `foldRight`, so I guess this is `foldRightWith` ?

distage/distage-framework/src/main/scala/izumi/distage/framework/services/ConfigLoader.scala
Caparow: > This parameter is redundant given ConfigLoader.map

I've miss this method ._.
Will update this

distage/distage-testkit-scalatest/src/test/scala/izumi/distage/testkit/distagesuite/DistageTestExample.scala
neko-kai: @Caparow 
So, by @pshirshov , this behavior is as expected... You may add a instantitation counter for `SetElement` and use it to skip the test with `assume` or just `assume(set.nonEmpty)`...

neko-kai:
@Caparow Could you please rebase on top of `develop`, not `series/zio-RC16` ?

distage/distage-testkit-core/src/main/scala/izumi/distage/testkit/services/dstest/SpecEnvironment.scala
Caparow: Yup, but `ConfigLoader#map` requires SpecEnvironment override to apply. Also, `moduleOverrides` (if I understood you properly) can not update config sources and point to another folder/file.

Caparow:
@neko-kai I suppose that all done ?

neko-kai (approved):

Issue #684, opened 2019-10-14
LogStage: add support for strings with .stripMargin


neko-kai:
This is a simple issue, suitable for beginners. https://github.com/7mind/izumi/pull/683 adds a failing test that shows `logstage` doesn't handle multiline string interpolations if they use `stripMargin`.  The error message prints the structure of the tree that should be supported:

```
Warning:(41, 44) Complex expression found as an input for a logger:
scala.Predef.augmentString(scala.StringContext.apply("This\\n           |is a\n           |multiline ", "").s(scala.Predef.ArrowAssoc[String](m).->[String]("message"))).stripMargin ; 
Select(Apply(Select(Select(Ident(scala), scala.Predef), TermName("augmentString")), List(
  Apply(Select(Apply(Select(Select(Ident(scala), scala.StringContext), TermName("apply")), List(Literal(Constant("This is a multiline ")), Literal(Constant("")))), TermName("s")), List(
  Apply(TypeApply(Select(Apply(TypeApply(Select(Select(Ident(scala), scala.Predef), TermName("ArrowAssoc")), List(TypeTree())), List(Ident(TermName("m")))), TermName("$minus$greater")), List(TypeTree())), 
  List(Literal(Constant("message")))))))), TermName("stripMargin")).
But Logstage expects you to use string interpolations instead, such as:
1) Simple variable:
   logger.info(s"My message: $argument")
2) Chain:
   logger.info(s"My message: ${call.method} ${access.value}")
3) Named expression:
   logger.info(s"My message: ${Some.expression -> "argname"}")
4) Invisible name expression:
   logger.info(s"My message: ${Some.expression -> "argname" -> null}")
5) De-camelcased name:
   logger.info(${camelCaseName-> ' '})
           |multiline ${m -> "message"}""".stripMargin
```

Adding a case that handles this tree in `ArgumentNameExtractionMacro` will fix the issue. After that the following test should pass:

```scala
"support .stripMargin" in {
  val m = "M E S S A G E"
  val message = Message {
    s"""This
       |is a
       |multiline ${m -> "message"}""".stripMargin
  }

  assert(message.template.parts == List("This\nis a\nmultiline ", ""))
  assert(message.args == List(LogArg(Seq("message"), m, hiddenName = false)))
}
```

neko-kai:
Fixed in #808 🎉

PR #808, opened 2020-01-02
LogStage: add stripMargin support.

VladPodilnyk:
Solution to #684

### DIFF
--- a/logstage/logstage-api/src/main/scala/izumi/logstage/macros/LogMessageMacro.scala
+++ b/logstage/logstage-api/src/main/scala/izumi/logstage/macros/LogMessageMacro.scala
@@ -148,6 +148,11 @@ object LogMessageMacro {
         val namedArgs = ArgumentNameExtractionMacro.recoverArgNames(c)(args.map(p => c.Expr(p)))
         reifyContext(c)(stringContext, namedArgs)
 
+      case Select(Apply(_, List(Apply(Select(stringContext@Apply(Select(Select(Ident(TermName("scala")), TermName("StringContext")), TermName("apply")), _), TermName("s")), args: List[c.Tree]))), TermName("stripMargin")) =>
+        val namedArgs = ArgumentNameExtractionMacro.recoverArgNames(c)(args.map(p => c.Expr(p)))
+        val sc = q"""StringContext($stringContext.parts.map(_.stripMargin): _*)"""
+        reifyContext(c)(sc, namedArgs)
+
       case Literal(c.universe.Constant(s)) =>
         val emptyArgs = reify(List.empty)
         val sc = q"StringContext(${s.toString})"
--- a/logstage/logstage-api/src/main/scala/izumi/logstage/macros/LogMessageMacro.scala
+++ b/logstage/logstage-api/src/main/scala/izumi/logstage/macros/LogMessageMacro.scala
@@ -150,7 +150,7 @@ object LogMessageMacro {
 
       case Select(Apply(_, List(Apply(Select(stringContext@Apply(Select(Select(Ident(TermName("scala")), TermName("StringContext")), TermName("apply")), _), TermName("s")), args: List[c.Tree]))), TermName("stripMargin")) =>
         val namedArgs = ArgumentNameExtractionMacro.recoverArgNames(c)(args.map(p => c.Expr(p)))
-        val sc = q"""StringContext($stringContext.parts.map(_.stripMargin): _*)"""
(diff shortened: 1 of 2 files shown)


logstage/logstage-api/src/main/scala/izumi/logstage/macros/LogMessageMacro.scala
```diff
@@ -148,6 +148,11 @@ object LogMessageMacro {
         val namedArgs = ArgumentNameExtractionMacro.recoverArgNames(c)(args.map(p => c.Expr(p)))
         reifyContext(c)(stringContext, namedArgs)
 
+      case Select(Apply(_, List(Apply(Select(stringContext@Apply(Select(Select(Ident(TermName("scala")), TermName("StringContext")), TermName("apply")), _), TermName("s")), args: List[c.Tree]))), TermName("stripMargin")) =>
+        val namedArgs = ArgumentNameExtractionMacro.recoverArgNames(c)(args.map(p => c.Expr(p)))
+        val sc = q"""StringContext($stringContext.parts.map(_.stripMargin): _*)"""
```
neko-kai: Better `_root_.scala.StringContext`, so that it won't break if user has a local variable named `StringContext`.

neko-kai (approved):
Nice! 👍

logstage/logstage-api/src/main/scala/izumi/logstage/macros/LogMessageMacro.scala
VladPodilnyk: Done.

neko-kai:
@VladPodilnyk 
Build failed for 2.13:
```
 should support .stripMargin *** FAILED *** (129 milliseconds)
[info]   List("This
[info]   is a
[info]   multiline (M E S S A G E,message)") did not equal List("This
[info]   is a
[info]   multiline ", "") (BasicLoggingTest.scala:44)
```

https://dev.azure.com/7mind/izumi/_build/results?buildId=1427&view=logs&j=daa08369-5d59-5ff0-43c1-a813c7da4ce0&t=55c0aa48-86d3-5ad0-e36d-5d778a28f034&l=3038

String interpolations work differently In 2.13 (they're transformed at compile-time into concatenations like `"abc" + "cba" + ("M E S S A G E", message)`. There's code in the macro that deals with this. You can switch IDEA to compile for 2.13 by switching comments here - https://github.com/7mind/izumi/blob/develop/project/Deps.sc#L147

VladPodilnyk:
@neko-kai, I've fixed the issue with `2.13`, but the solution is a bit ugly, especially this https://github.com/7mind/izumi/pull/808/files#diff-bf805d28850a4b89582dc3d4bfc063b5R53 - bad for performance.
I'll try to come up with smth better tomorrow.

logstage/logstage-api/src/main/scala/izumi/logstage/macros/LogMessageMacro.scala
```diff
@@ -154,13 +156,18 @@ object LogMessageMacro {
         val sc = q"""_root_.scala.StringContext($stringContext.parts.map(_.stripMargin): _*)"""
         reifyContext(c)(sc, namedArgs)
 
-       // support .stripMargin in scala 2.13
+      // support .stripMargin in scala 2.13
       case Select(Apply(_, List(Typed(tree, _))), TermName("stripMargin")) =>
-        processExpr(c)(tree)
+        tree match {
+          case PlusExtractor(lst) =>
```
neko-kai: @VladPodilnyk 
Hmm, I'm not sure this will always match (what if there are no interpolations? e.g. `"abc".stripMargin`, there probably won't be any pluses)
IMHO, using tailrec here is fine, and I don't see any performance issues - `@tailrec` recursion is essentially free, there isn't even a method call - you may move `isMultiline` argument up to `processExpr` and just recurse againt with `processExpr(c)(tree, true)`

VladPodilnyk: Done. @neko-kai, I'll merge this?

neko-kai: @VladPodilnyk Yep 👍

neko-kai: @VladPodilnyk Actually, could you add a test for single-line string with stripMargin & no-interpolations string with stripMargin, then merge? Thanks!

neko-kai (approved):

PR #819, opened 2020-01-13
feature/docker-networks

Caparow:


distage/distage-framework-docker/src/main/scala/izumi/distage/docker/DockerClientWrapper.scala
```diff
@@ -30,21 +30,25 @@ class DockerClientWrapper[F[_]]
...
+        } finally {
+          client
+            .removeContainerCmd(container.name)
+            .withForce(true)
+            .exec()
+            .discard()
+        }
       }
+    } {
+      e => F.maybeSuspend(logger.warn(s"Got failure during container destroy ${e.getMessage -> "message"}"))
```
neko-kai, line 51: @Caparow Better add a `catch` clause to try-finally above. `e` in `definitelyRecover` is the entire ZIO Cause, so `e.getMessage` could be unnecessarily verbose.

distage/distage-framework-docker/src/main/scala/izumi/distage/docker/DockerContainer.scala
```diff
@@ -245,6 +247,26 @@ object DockerContainer {
...
               logger.debug(s"Created $container from ${config.image}...")
+
+              if (config.networks.nonEmpty) {
+                logger.debug(s"Going to attach container ${res.getId -> "id"} to ${config.networks -> "networks"}")
+
+                val existedNetworks: List[Network] = client.listNetworksCmd().exec().asScala.toList
+
+                config.networks.map {
+                  network =>
+                    existedNetworks.find(_.getName == network.name).fold {
```
neko-kai, line 258: Maybe we should create networks with random name and manage reuse same as we do with containers? /cc @pshirshov

distage/distage-framework-docker/src/main/scala/izumi/distage/docker/DockerClientWrapper.scala
Caparow: Yup
Will update

distage/distage-framework-docker/src/main/scala/izumi/distage/docker/DockerContainer.scala
Caparow: @neko-kai I've thought of creation resource managed networks. To bind them via weak sets. But there was trouble that in such case we will need to add network binding, add a dependency to the container (to strengthen reference, I thought of method like `dependsOnDocker`) and modify container config to get an updated list of networks. Which is a little bit overcomplicated then just add networks in one place.

Caparow: >  modify container config to get an updated list of networks

Only if we would have any chance to do this during `dependsOnNetwork` method (like add a dependency to `ProvideMagnet` and modify config before instantiation.)

Issue #935, opened 2020-03-10
Add a bundled ContainerDef for Kafka & Zookeeper in distage-framework-docker


neko-kai:

0lejk4:
Hi, I managed to create working solution could you take a look, so we can decide how can this be added to distage. The solution is the same as in test containers-java, This one can be tweaked to use external zookeeper.
```
import izumi.distage.docker.ContainerDef
import izumi.distage.docker.Docker.{ContainerConfig, DockerPort}

object KafkaDocker extends ContainerDef {

  implicit class KafkaContainer(container: Container) {
    def bootstrapServer: String = {
      val address = container.availablePorts(KafkaDocker.primaryPort).head
      s"PLAINTEXT://${address.hostV4}:${address.port}"
    }
  }

  val primaryPort: DockerPort = DockerPort.TCP(9093)
  val StarterScript = "/testcontainers_start.sh"

  override def config: Config =
    ContainerConfig(
      image = "confluentinc/cp-kafka:5.2.1",
      cmd = Seq("sh", "-c", s"while [ ! -f $StarterScript ]; do sleep 0.05; done; $StarterScript"),
      env = Map(
        "KAFKA_ZOOKEEPER_CONNECT" -> "localhost:2181",
        "KAFKA_LISTENERS" -> "PLAINTEXT://0.0.0.0:9093,BROKER://0.0.0.0:9092",
        "KAFKA_LISTENER_SECURITY_PROTOCOL_MAP" -> "BROKER:PLAINTEXT,PLAINTEXT:PLAINTEXT",
        "KAFKA_INTER_BROKER_LISTENER_NAME" -> "BROKER",
        "KAFKA_BROKER_ID" -> "1",
        "KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR" -> "1",
        "KAFKA_OFFSETS_TOPIC_NUM_PARTITIONS" -> "1",
        "KAFKA_AUTO_CREATE_TOPICS_ENABLE" -> "true",
        "KAFKA_LOG_FLUSH_INTERVAL_MESSAGES" -> Long.MaxValue.toString,
        "KAFKA_GROUP_INITIAL_REBALANCE_DELAY_MS" -> "0",
      ),
      ports = Seq(primaryPort),
      reuse = false,
    )
}

//Some actions I need to start everything nicely
  implicit class DockerClientOps[F[_]](wrapper: DockerClientWrapper[F]) {
[...]

PR #939, opened 2020-03-12
Bundled containers for ZK, Kafka, C*

Caparow:
Issues:
#936 
#935

### DIFF
--- a/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/examples/CassandraDocker.scala
+++ b/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/examples/CassandraDocker.scala
@@ -1,0 +1,27 @@
+package izumi.distage.docker.examples
+
+import izumi.distage.docker.ContainerDef
+import izumi.distage.docker.Docker.{ContainerConfig, DockerPort}
+import izumi.distage.model.definition.ModuleDef
+import izumi.fundamentals.reflection.Tags.TagK
+
+object CassandraDocker extends ContainerDef {
+  val primaryPort: DockerPort = DockerPort.TCP(9042)
+
+  override def config: Config = {
+    ContainerConfig(
+      image = "cassandra:3.0.16",
+      ports = Seq(primaryPort),
+      env = Map(
+        "CASSANDRA_VERSION" -> "3.0.16",
+        "CASSANDRA_CONFIG" -> "/etc/cassandra"
+      )
+    )
+  }
+}
+
+class CassandraDockerModule[F[_]: TagK] extends ModuleDef {
+  make[CassandraDocker.Container].fromResource {
+    CassandraDocker.make[F]
+  }
+}
--- a/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/examples/CassandraDocker.scala
+++ b/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/examples/CassandraDocker.scala
@@ -10,7 +10,7 @@ object CassandraDocker extends ContainerDef {
 
   override def config: Config = {
     ContainerConfig(
-      image = "cassandra:latest",
+      image = "cassandra:3.11.6",
       ports = Seq(primaryPort)
     )
   }
--- a/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/examples/CassandraDocker.scala
+++ b/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/examples/CassandraDocker.scala
@@ -10,12 +10,8 @@ object CassandraDocker extends ContainerDef {
 
   override def config: Config = {
     ContainerConfig(
-      image = "cassandra:3.0.16",
-      ports = Seq(primaryPort),
-      env = Map(
-        "CASSANDRA_VERSION" -> "3.0.16",
-        "CASSANDRA_CONFIG" -> "/etc/cassandra"
-      )
+      image = "cassandra:latest",
+      ports = Seq(primaryPort)
     )
   }
 }
--- a/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/examples/DynamoDocker.scala
+++ b/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/examples/DynamoDocker.scala
@@ -9,7 +9,7 @@ object DynamoDocker extends ContainerDef {
 
   override def config: Config = {
     ContainerConfig(
-      image = "amazon/dynamodb-local:1.12.0",
+      image = "amazon/dynamodb-local:latest",
       ports = Seq(primaryPort),
     )
   }
--- a/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/examples/DynamoDocker.scala
+++ b/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/examples/DynamoDocker.scala
@@ -9,7 +9,7 @@ object DynamoDocker extends ContainerDef {
 
   override def config: Config = {
     ContainerConfig(
-      image = "amazon/dynamodb-local:latest",
+      image = "amazon/dynamodb-local:1.12.0",
       ports = Seq(primaryPort),
     )
   }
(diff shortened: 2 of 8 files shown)


0lejk4:
I may be incorrect but isn't such Kafka/Zookeeper setup disallows creation of multiple containers? Because you assign zookeeper container a name there can be only one instance with the same name. And as far as I know in one zookeeper instance there can only be one Kafka instance. The second one you'll try to start will fail. It would be nice if it was possible to start them with random names and on random networks, so multiple parallel integration tests would be possible.

Caparow:
@0lejk4 I've updated the solution.
Now each run will be created new random ZK and Kafka (under one network since containers linked to each other by names and there is no reason to spawn new networks). 
But, I do not understand why you want to use an independent container for each test if you could use random topics (or ZK directories if you are using it directly). Also, with DI we are binding singletons, if you want multiple independent containers you should bind them with different `DIKey`s.
E.g. 
```scala
make[KafkaDocker.Container].named("test-kafka-1").fromResource{...}
```

Please, @neko-kai correct me if I am wrong. Also, I don't quite like current bindings, but I cannot combine `ProviderMagnets` to use this like this `(zk: Zookeeper) => new Kafka(zk).make[F]` (it will return ProviderMagnet[ProviderMagnet[...]).

distage/distage-framework-docker/src/main/scala/izumi/distage/docker/examples/CassandraDocker.scala
```diff
@@ -0,0 +1,27 @@
...
+import izumi.distage.docker.Docker.{ContainerConfig, DockerPort}
+import izumi.distage.model.definition.ModuleDef
+import izumi.fundamentals.reflection.Tags.TagK
+
+object CassandraDocker extends ContainerDef {
+  val primaryPort: DockerPort = DockerPort.TCP(9042)
+
+  override def config: Config = {
+    ContainerConfig(
+      image = "cassandra:3.0.16",
```
neko-kai: Seems like the latest version is `3.11.6` ? https://hub.docker.com/_/cassandra/ Since these are just bundled examples, we may as well ship the latest versions, if someone needs an older one they can make their own copy

distage/distage-framework-docker/src/main/scala/izumi/distage/docker/examples/KafkaDocker.scala
```diff
@@ -0,0 +1,34 @@
...
+import izumi.distage.docker.Docker.{ContainerConfig, DockerPort}
+import izumi.distage.docker.{ContainerDef, DockerClientWrapper, DockerContainer}
+import izumi.distage.model.effect.{DIEffect, DIEffectAsync}
+import izumi.logstage.api.IzLogger
+
+object KafkaDocker extends ContainerDef {
+  val primaryPort: DockerPort.TCP = DockerPort.TCP(9092)
+  override def config: Config = {
+    ContainerConfig(
+      image = "wurstmeister/kafka:1.0.0",
```
neko-kai: Seems like the latest version is `2.12-2.4.0` , https://github.com/wurstmeister/kafka-docker#tags-and-releases

distage/distage-framework-docker/src/main/scala/izumi/distage/docker/examples/ZookkeeperDocker.scala
```diff
@@ -0,0 +1,29 @@
...
+import izumi.distage.docker.Docker.{ContainerConfig, ContainerNetwork, DockerPort}
+import izumi.distage.docker.{ContainerDef, DockerClientWrapper, DockerContainer}
+import izumi.distage.model.effect.{DIEffect, DIEffectAsync}
+import izumi.logstage.api.IzLogger
+
+object ZookeeperDocker extends ContainerDef {
+  val zookeeperNetwork: ContainerNetwork = ContainerNetwork("zookeeper-network")
+  override def config: Config = {
+    ContainerConfig(
+      image = "zookeeper:3.4.13",
```
neko-kai: Seems like the latest version is `3.5.17` https://hub.docker.com/_/zookeeper/

distage/distage-framework-docker/src/main/scala/izumi/distage/docker/examples/ZookkeeperDocker.scala
```diff
@@ -1,23 +1,29 @@
...
+      reuse = false
     )
   }
 }
 
 class ZookeeperDockerModule[F[_]: TagK] extends ModuleDef {
   make[ZookeeperDocker.Container].fromResource {
-    ZookeeperDocker.make[F]
+    (wr: DockerClientWrapper[F], log: IzLogger, eff: DIEffect[F], effAsync: DIEffectAsync[F]) =>
+      val randomName = s"zookeeper-${UUID.randomUUID().toString.take(8)}"
```
neko-kai: Hmm, the names are always random already if you put `name = None`, i think this is not necessary...

distage/distage-framework-docker/src/main/scala/izumi/distage/docker/examples/ZookkeeperDocker.scala
```diff
@@ -1,23 +1,29 @@
...
+
 import distage.{ModuleDef, TagK}
-import izumi.distage.docker.ContainerDef
 import izumi.distage.docker.Docker.{ContainerConfig, ContainerNetwork, DockerPort}
+import izumi.distage.docker.{ContainerDef, DockerClientWrapper, DockerContainer}
+import izumi.distage.model.effect.{DIEffect, DIEffectAsync}
+import izumi.logstage.api.IzLogger
 
 object ZookeeperDocker extends ContainerDef {
   val zookeeperNetwork: ContainerNetwork = ContainerNetwork("zookeeper-network")
```
neko-kai, line 8: We will add randomization for network names in `0.10.2` too, for now that's okay

neko-kai:
> Please, @neko-kai correct me if I am wrong. Also, I don't quite like current bindings, but I cannot combine ProviderMagnets to use this like this (zk: Zookeeper) => new Kafka(zk).make[F] (it will return ProviderMagnet[ProviderMagnet[...]).

@Caparow Yes, there's no way to combine providers monadically (in that case you could only get some of dependencies after applying the function, but this case is impossible to support with our static planning paradigm)

But I think it's not necessary in this case. I think you don't need to randomize names, because they're already randomized. Setting `reuse = false` is enough to ensure new containers get spawned for both ZK and Kafka

distage/distage-framework-docker/src/main/scala/izumi/distage/docker/examples/ZookkeeperDocker.scala
Caparow: Yup, but with this, we can access ZK container from Kafka's container via aliases.
Docker does not create name alias for containers with random names.

Caparow:
@neko-kai all containers now using the latest tags. Also, now we will fetch hostnames from container inspections and may use names generated by Docker. I suppose we can merge this PR since all done?
@0lejk4 you may use these examples to build your own composed docker test environments.

distage/distage-framework-docker/src/main/scala/izumi/distage/docker/examples/DynamoDocker.scala
```diff
@@ -9,7 +9,7 @@ object DynamoDocker extends ContainerDef {
 
   override def config: Config = {
     ContainerConfig(
-      image = "amazon/dynamodb-local:1.12.0",
+      image = "amazon/dynamodb-local:latest",
```
neko-kai: Don't explicitly use `latest` because there may be unexpected upstream regressions on latest version, e.g. `postgres:latest` no longer just works without additional configuration right now:
```
[~] % docker run postgres:latest
Error: Database is uninitialized and superuser password is not specified.
       You must specify POSTGRES_PASSWORD for the superuser. Use
       "-e POSTGRES_PASSWORD=password" to set it in "docker run".

       You may also use POSTGRES_HOST_AUTH_METHOD=trust to allow all connections
       without a password. This is *not* recommended. See PostgreSQL
       documentation about "trust":
       https://www.postgresql.org/docs/current/auth-trust.html
```

So we may want to use latest known good versions, but not just blindly `latest`

Caparow: Done :+1:

neko-kai:
Merged!

neko-kai:
@0lejk4 Could you please test if these containers can work for you once a new release is cut? If not, we can figure out a solution if you can provide enough information about the issues you have. Thanks!

0lejk4:
@neko-kai Yes, looking forward for release. Thank you for PR @Caparow

PR #942, opened 2020-03-13
Feature/bio3 2

Caparow:


fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/instances/BIOApplicative3.scala
```diff
@@ -4,13 +4,13 @@ trait BIOApplicative3[F[-_, +_, +_]] extends BIOBifunctor3[F] {
   def pure[A](a: A): F[Any, Nothing, A]
 
   /** execute two operations in order, map their results */
-  def map2[R, E, A, B, C](firstOp: F[R, E, A], secondOp: => F[R, E, B])(f: (A, B) => C): F[R, E, C]
+  def map2[R, R1 <: R, E, E1 >: E, A, B, C](firstOp: F[R, E, A], secondOp: => F[R1, E1, B])(f: (A, B) => C): F[R1, E1, C]
```
neko-kai: @Caparow 
This is redundant because both arguments are widened to their common upper-bound already due to being on one parameter list. e.g. `map2(a: _Nothing, b: _Any) -> E=Any`
Variant variables are only required in syntax or with split parameter lists

PR #987, opened 2020-04-05
proper concatenation of Message

### LINKED ISSUE #983: logstage: concatenation of Message results in trimmed spaces at the beginning
```scala
import izumi.logstage.api.IzLogger
import izumi.logstage.api.Log.Level.Debug
import izumi.logstage.api.Log.Message
import izumi.logstage.sink.ConsoleSink

val log = IzLogger(Debug, Seq(ConsoleSink.SimpleConsoleSink))
val a = 0
val b = 1
val c = 2
val d = 3

val message = Message(s"$a") +
  Message(s" $b") +
  Message(s" $c") +
  Message(s" $d")

log.log(Debug)(message)
```
Expected output: `a=0 b=1 c=2 d=3`
Actual output: `a=0b=1 c=2d=3 `

There's no space between arg groups of two, this pattern continues for more concatenations. Adding parenthesis doesn't help.

Putting the spaces to the end
```scala
val message = Message(s"$a ") +
  Message(s"$b ") +
  Message(s"$c ") +
  Message(s"$d ")
```

results in `a=0 b=1c=2 d=3`

Having spaces around leads to the expected output:
```scala
val message = Message(s"$a ") +
  Message(s" $b ") +
  Message(s" $c ") +
  Message(s" $d")
```
-> `a=0 b=1 c=2 d=3 ` (except there seems to be an extra space after d)

I guess it's an issue around the concatenation of the StringContexts, but I couldn't quite nail it down.

Related: https://github.com/7mind/izumi/issues/975

an-tex:
fixes https://github.com/7mind/izumi/issues/983

### DIFF
--- a/logstage/logstage-core/src/main/scala/izumi/logstage/api/Log.scala
+++ b/logstage/logstage-core/src/main/scala/izumi/logstage/api/Log.scala
@@ -157,7 +157,17 @@ object Log {
   final case class Message(template: StringContext, args: LogContext) {
     def argsMap: Map[String, Set[Any]] = args.map(kv => (kv.name, kv.value)).toMultimap
 
-    def ++(that: Message): Message = Message(StringContext(template.parts ++ that.template.parts: _*), args ++ that.args)
+    def ++(that: Message): Message = {
+      // this ain't pretty but avoids multiple iterations on the same parts collection compared to using .init/.tail
+      val (thisInit, thisTails) = template.parts.splitAt(template.parts.length - 1)
+      val thisTail = thisTails.head
+
+      val (thatHeads, thatTail) = that.template.parts.splitAt(1)
+      val thatHead = thatHeads.head
+
+      val parts = (thisInit :+ (thisTail + thatHead)) ++ thatTail
+      Message(StringContext(parts: _ *), args ++ that.args)
+    }
     def +(that: Message): Message = ++(that)
   }
   object Message {
--- a/logstage/logstage-core/src/main/scala/izumi/logstage/api/Log.scala
+++ b/logstage/logstage-core/src/main/scala/izumi/logstage/api/Log.scala
@@ -158,15 +158,19 @@ object Log {
     def argsMap: Map[String, Set[Any]] = args.map(kv => (kv.name, kv.value)).toMultimap
 
     def ++(that: Message): Message = {
-      // this ain't pretty but avoids multiple iterations on the same parts collection compared to using .init/.tail
-      val (thisInit, thisTails) = template.parts.splitAt(template.parts.length - 1)
-      val thisTail = thisTails.head
+      if (that.template.parts.isEmpty) this
+      else if (template.parts.isEmpty) that
+      else {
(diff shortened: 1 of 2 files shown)


logstage/logstage-core/src/main/scala/izumi/logstage/api/Log.scala
```diff
@@ -157,7 +157,17 @@ object Log {
   final case class Message(template: StringContext, args: LogContext) {
     def argsMap: Map[String, Set[Any]] = args.map(kv => (kv.name, kv.value)).toMultimap
 
-    def ++(that: Message): Message = Message(StringContext(template.parts ++ that.template.parts: _*), args ++ that.args)
+    def ++(that: Message): Message = {
+      // this ain't pretty but avoids multiple iterations on the same parts collection compared to using .init/.tail
+      val (thisInit, thisTails) = template.parts.splitAt(template.parts.length - 1)
+      val thisTail = thisTails.head
```
neko-kai: @an-tex Could you please make it work if the left Message is empty (`Message(StringContext(), Nil)`) or consists of one part (`Message("")`), and add a test for that?

logstage/logstage-core/src/main/scala/izumi/logstage/api/Log.scala
```diff
@@ -157,7 +157,17 @@ object Log {
...
+
+      val (thatHeads, thatTail) = that.template.parts.splitAt(1)
+      val thatHead = thatHeads.head
```
neko-kai: @an-tex Could you please make it work if the right Message is empty (`Message(StringContext(), Nil)`) or consists of one part (`Message("")`), and add a test for that?

logstage/logstage-core/src/main/scala/izumi/logstage/api/Log.scala
an-tex: Is `StringContext()` a valid StringContext? Can't see how StringContext.checkLengths(args) could NOT throw an exception for any arguments. An empty Message is also defined as `def empty: Message = Message(new StringContext(""), Nil)`. The code already works for `Message.empty` and `Message("")` but I've added tests documenting it.

neko-kai: @an-tex Yeah, `checkLengths` is not called in StringContext constructor, so nothing stops you from constructing a `Message(StringContext(), Nil)` manually - because it's a value that can legally happen I think it shouldn't throw in `++`. Or I guess you could add a call `template.checkLengths(args)` in Message constructor which would forbid such a value from appearing.

an-tex: Fair point. As throwing an exception in the `Message` constructor might break existing functionality for people already using an empty StringContext for whatever reason I've opted in to just handle  that case in the `++`

neko-kai: Thanks!

neko-kai (approved):

neko-kai:
@an-tex Thank you so much! 🙏

PR #996, opened 2020-04-08
Monix BIO instances

VladPodilnyk:


fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/BIOFiber.scala
```diff
@@ -20,11 +21,19 @@ object BIOFiber3 {
...
   }
+
 }
 
 object BIOFiber {
   @inline def fromZIO[E, A](f: Fiber[E, A]): BIOFiber3[ZIO, E, A] = BIOFiber3.fromZIO(f)
+  @inline def fromMonix[E, A](f: bio.Fiber[E, A]): BIOFiber[bio.IO, E, A] = new BIOFiber[bio.IO, E, A] {
+    override def join: bio.IO[E, A] = f.join
+    // I could perform  only cancel
+    override def observe: bio.IO[Nothing, BIOExit[E, A]] = ???
```
neko-kai: @VladPodilnyk I think `observe` may be implementable as `fiber.join.sandboxBIOExit`

project/Deps.sc
```diff
@@ -373,9 +375,9 @@ object Izumi {
       ),
       Artifact(
         name = Projects.fundamentals.bio,
-        libs = allMonadsOptional ++ Seq(
-            scala_java_time in Scope.Test.js
-          ),
+        libs = allMonadsOptional ++ Seq(monix_bio in Scope.Compile.all) ++ Seq(
```
neko-kai: Could you make monix `Optional`, like zio?

VladPodilnyk: Done.

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/BIOFiber.scala
VladPodilnyk: Done.

project/plugins.sbt
```diff
@@ -1,7 +1,7 @@
 
 ////////////////////////////////////////////////////////////////////////////////
 
-addSbtPlugin("io.7mind.izumi.sbt" % "sbt-izumi" % "0.0.61")
+addSbtPlugin("io.7mind.izumi.sbt" % "sbt-izumi" % "0.0.60")
```
neko-kai: artefact

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/BIOFork3.scala
```diff
@@ -7,9 +7,10 @@ trait BIOFork3[F[-_, +_, +_]] extends BIOForkInstances {
 }
 
 private[bio] sealed trait BIOForkInstances
+
 object BIOForkInstances extends LowPriorityBIOForkInstances {
   implicit val BIOForkZio: BIOFork3[ZIO] = new BIOFork3[ZIO] {
-    override def fork[R, E, A](f: ZIO[R, E, A]): ZIO[R, Nothing, BIOFiber3[ZIO, E, A]] =
+    @inline override def fork[R, E, A](f: ZIO[R, E, A]): ZIO[R, Nothing, BIOFiber3[ZIO, E, A]] =
```
neko-kai: Please do not use inline here as there is little chance for this method to be inlined (there must be a direct call using the val `BIOForkZIO.fork` - which is unlikely as most calls are polymorphic against some parameter `bioFork: BIOFork[F]` - in addition I'm not sure if anonymous classes in `val`s can participate in inlining, possibly you would need to make this an `object` for inline to trigger)

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/impl/BIOAsyncMonix.scala
```diff
@@ -0,0 +1,132 @@
...
+import izumi.functional.bio.{BIOAsync, BIOExit, BIOFiber}
+import monix.bio
+import monix.bio.{Cause, IO, Task}
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.Try
+
+object BIOAsyncMonix extends BIOAsyncMonix
+
+class BIOAsyncMonix extends BIOAsync[IO] {
```
neko-kai, line 14: Do you think `BIOTemporal` is implementable for monix-bio? (or monix-bio + `cats.effect.Clock`?)

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/impl/BIOAsyncMonix.scala
```diff
@@ -0,0 +1,132 @@
...
+
+import izumi.functional.bio.BIOExit.MonixExit._
+import izumi.functional.bio.{BIOAsync, BIOExit, BIOFiber}
+import monix.bio
+import monix.bio.{Cause, IO, Task}
+
+import scala.concurrent.{ExecutionContext, Future}
+import scala.util.Try
+
+object BIOAsyncMonix extends BIOAsyncMonix
```
neko-kai, line 12: I think this also needs to be exposed as an implicit (in `BIORootInstances`)

neko-kai (commented):
I think this is almost totally ready

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/impl/BIOAsyncMonix.scala
VladPodilnyk: Done

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/impl/BIOAsyncMonix.scala
VladPodilnyk: I don't thinks so. I haven't spotted any functionality to repeat/retry using Schedulers. Um, maybe because `cats-retry` exists, so they though the one could use it with monix bio.

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/BIOFork3.scala
VladPodilnyk: Understood, thanks.

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/BIORoot.scala
```diff
@@ -69,6 +70,7 @@ sealed trait BIORootInstancesLowPriority5 extends BIORootInstancesLowPriority6 {
 }
 
 sealed trait BIORootInstancesLowPriority6 {
+  @inline implicit final def BIOMonix: Predefined.Of[BIOAsync[bio.IO]] = Predefined(BIOAsyncMonix)
```
neko-kai: I think you should move the BIOConvert3To2 instance ta lower priority trait `BIORootInstancesLowPriority7` now, so that these two instances do not interfere

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/impl/BIOAsyncMonix.scala
neko-kai: @VladPodilnyk `BIOTemporal#retryOrElse` does not require Schedule though, it's easily implementable manually (you should decrease the remaning duration after each failed try and execute `orElse` once the alloted time for retrying is up)

neko-kai:
@VladPodilnyk 
Also we now need a test that `Optional` dependency on monix works, correctly - that is, implicit searches still work and `BIOConvert3To2` instance is accessible. Check out `OptionalDependencyTest` in distage-config/test, you may add a test that uses `BIOConvert3To2` instance there to check that scalac is not broken when monix-bio is not a dependency, e.g. a test like this:

```scala
def threeTo2[FR[-_, +_, +_]](implicit FR: BIO3[FR]): FR[Any, Nothing, Unit] = {
  val F: BIO[F[Any, +?, +?]] = implicitly // must use `BIOConvert3To2` instance to convert FR -> F
  F.unit
}
```

distage/distage-framework/src/main/scala/izumi/distage/effect/modules/MonixDIEffectModule.scala
```diff
@@ -0,0 +1,11 @@
+package izumi.distage.effect.modules
+
+import distage.ModuleDef
+import izumi.distage.model.effect.DIEffectRunner
+import monix.bio
+
+object MonixDIEffectModule extends MonixDIEffectModule
+
+trait MonixDIEffectModule extends ModuleDef {
+  addImplicit[DIEffectRunner[bio.Task]]
```
neko-kai: @VladPodilnyk  Just DIEffectRunner is not enough to run distage on monix, at least `DIEffect` & `DIEffectAsync` are also required (these can just use `cats` instance). Also the module should include all the cats instances, like `CatsDIEffectModule` (There's actually already a polymorphic implementation for `DIEffectModule` for Monix - https://github.com/BlueBrain/nexus/blob/cb8b4a587acbef7d0ad9d77240e4d6c30a6b20fe/cli/src/main/scala/ch/epfl/bluebrain/nexus/cli/EffectModule.scala we may reuse it and make `CatsDIEffectModule` polymorphic)

VladPodilnyk: Yup, I know that I need `DIEffect` and `DIEffectAsync` (that is the reason why the tests are broken). I am just thinking of reusing `cats` instance and you kindly pointed me in the right direction.

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/impl/BIOAsyncMonix.scala
VladPodilnyk: Um, do u mean decrease by effect execution time? Otherwise, it's just does not make any sense, 'cuz it won't be related to a time, but to number of retries, am I right?

neko-kai: @VladPodilnyk Yeah

distage/distage-framework/src/main/scala/izumi/distage/effect/modules/CatsDIEffectModule.scala
```diff
@@ -39,12 +17,10 @@ trait CatsDIEffectModule extends ModuleDef {
   make[ContextShift[IO]].from((_: PublicIOApp).contextShift)
   make[Timer[IO]].from((_: PublicIOApp).timer)
 
-  make[PublicIOApp].from {
-    new PublicIOApp { override def run(args: List[String]): IO[ExitCode] = IO.pure(ExitCode(0)) }
-  }
+  implicit val cs: ContextShift[IO] = IO.contextShift(ExecutionContext.global)
```
neko-kai: @VladPodilnyk ContextShift is already bound in DI, why bind it at top-level? (And also this uses a different ContextShift not from DI, which results in having multiple ContextShift's bound to different thread pools in the app)

distage/distage-framework/src/main/scala/izumi/distage/effect/modules/PolymorphicCatsDIEffectModule.scala
```diff
@@ -0,0 +1,46 @@
...
+import distage.{ModuleDef, TagK}
+import izumi.distage.model.effect.{DIApplicative, DIEffect, DIEffectAsync, DIEffectRunner}
+import izumi.functional.mono.SyncSafe
+import monix.bio.Task
+
+/**
+  * Module definition that binds effect TC instances for an arbitrary F[_].
+  *
+  */
+class PolymorphicCatsDIEffectModule[F[_]: ConcurrentEffect: TagK] extends ModuleDef {
```
neko-kai: > `: ConcurrentEffect`

I think it's a bad idea to require ConcurrentEffect to be an external value already created somehow, instead of expecting it to come from DI (where it can be configured)

So, instead of requiring this instance as a value, you may require it in each binding separately, e.g.

instead of 
```scala
  addImplicit[Functor[F]]
  addImplicit[Applicative[F]]
  addImplicit[Monad[F]]
```
->
```scala
  make[Functor[F]].using[ConcurrentEffect[F]]
  make[Applicative[F]].using[ConcurrentEffect[F]]
  make[Monad[F]].using[ConcurrentEffect[F]]
```

and instead of

```scala
  addImplicit[DIEffect[F]]
```
->
```scala
  make[DIEffect[F]].from { implicit F: Sync[F] => DIEffect.fromCatsEffect }
```

distage/distage-framework/src/main/scala/izumi/distage/effect/modules/PolymorphicCatsDIEffectModule.scala
```diff
@@ -0,0 +1,46 @@
...
+  implicit private def diEffectRunner: DIEffectRunner[F] =
```
neko-kai: I think it may be a good idea to add this instance `DIEffectRunner` companion object instead of making it private

distage/distage-framework/src/main/scala/izumi/distage/effect/modules/PolymorphicCatsDIEffectModule.scala
```diff
@@ -0,0 +1,46 @@
...
+    new DIEffectRunner[F] {
+      override def run[A](f: => F[A]): A = ConcurrentEffect[F].toIO(f).unsafeRunSync()
```
neko-kai: In this case `Effect[F]` is sufficient (esp. if you extract it into a companion object instance)

distage/distage-framework/src/main/scala/izumi/distage/effect/modules/PolymorphicCatsDIEffectModule.scala
```diff
@@ -0,0 +1,46 @@
...
+      DIEffectAsync[F]
+  }
+
+  addImplicit[Functor[F]]
+  addImplicit[Applicative[F]]
+  addImplicit[Monad[F]]
+  addImplicit[MonadError[F, Throwable]]
+  addImplicit[Bracket[F, Throwable]]
+  addImplicit[Sync[F]]
+  addImplicit[Async[F]]
```
neko-kai: forgot `Effect[F]`

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/impl/BIOTemporalMonix.scala
```diff
@@ -0,0 +1,27 @@
+package izumi.functional.bio.impl
+
+import izumi.functional.bio.{BIOTemporal, Clock2}
+import monix.bio.IO
+
+import scala.concurrent.duration._
+
+class BIOTemporalMonix(clock: Clock2[IO]) extends BIOAsyncMonix with BIOTemporal[IO] {
```
neko-kai, line 8: 👍

distage/distage-core-api/.jvm/src/main/scala/izumi/distage/model/effect/DIEffectRunner.scala
```diff
@@ -12,6 +12,11 @@ trait DIEffectRunner[F[_]] {
 object DIEffectRunner {
   def apply[F[_]: DIEffectRunner]: DIEffectRunner[F] = implicitly
 
+  def catsAndMonix[F[_]: ConcurrentEffect]: DIEffectRunner[F] =
```
neko-kai: `Effect` is enough

project/plugins.sbt
```diff
@@ -1,7 +1,7 @@
 
 ////////////////////////////////////////////////////////////////////////////////
 
-addSbtPlugin("io.7mind.izumi.sbt" % "sbt-izumi" % "0.0.60")
+addSbtPlugin("io.7mind.izumi.sbt" % "sbt-izumi" % "0.0.61")
```
VladPodilnyk: I guess this is redundant

neko-kai: @VladPodilnyk Fixed in https://github.com/7mind/izumi/pull/1189

VladPodilnyk: I've rebased the branch.

project/Deps.sc
```diff
@@ -531,7 +533,7 @@ object Izumi {
     artifacts = Seq(
       Artifact(
         name = Projects.docs.microsite,
-        libs = (cats_all ++ zio_all ++ http4s_all ++ doobie).map(_ in Scope.Compile.all),
+        libs = (cats_all ++ zio_all ++ http4s_all ++ doobie ++ List(monix_bio)).map(_ in Scope.Compile.all),
```
neko-kai: @VladPodilnyk Hmm, why does microsite require monix-bio now?

VladPodilnyk: @neko-kai, microsite build fails because mdoc couldn't find Monix's types  in the classpath

neko-kai: @VladPodilnyk Do you know what causes it to try to find them?

neko-kai: They shouldn't be referenced anywhere...

VladPodilnyk: I catch the  following error
```
info: Compiling 35 files to /Users/vpodilnyk/work/izumi/doc/microsite/target/mdoc
error: /Users/vpodilnyk/work/izumi/doc/microsite/src/main/tut/bio/00_bio.md:21:3: Symbol 'type monix.bio.IO' is missing from the classpath.
This symbol is required by 'method izumi.functional.bio.BIORootInstancesLowPriority6.BIOMonix'.
Make sure that type IO is in your classpath and check for conflicting dependencies with `-Ylog-classpath`.
A full rebuild may help if 'BIORootInstancesLowPriority6.class' was compiled against an incompatible version of monix.bio.
  F.mkRef(0)
  ^
info: Compiled in 4.59s (1 error)

```

VladPodilnyk: I've done `sbt doc/makeSite`

VladPodilnyk: So, basically,  I just followed by compiler suggestion and add  `monix` as dependency.

neko-kai: @VladPodilnyk But that means that optional dependency is not actually optional

neko-kai: @VladPodilnyk Could you add the method that produces the error to OptionalDependencyTest and check what happens there?

VladPodilnyk: @neko-kai, this test passes, but I expect to catch `NullPointerException` here (like with `BIOConvert3to2` described below):
```scala
 "test " in {
    def adder[F[+_, +_]: BIOMonad: BIOPrimitives](i: Int) = BIOPrimitives[F].mkRef(0).flatMap(ref => ref.update(_ + i) *> ref.get)

    implicit val bioMonad: BIOMonad[Identity2] = null
    implicit val primitives: BIOPrimitives[Identity2] = null
    assert(adder[Identity2](0) == 0)
  }
```

BIOConver3to2 test:
```scala
def threeTo2[FR[-_, +_, +_]](implicit FR: BIO3[FR]): FR[Any, Nothing, Unit] = {
      val F: BIO[FR[Any, +?, +?]] = implicitly // must use `BIOConvert3To2` instance to convert FR -> F
      F.unit
    }

    type EitherFn[-R, +E, +A] = R => Either[E, A]
    implicit val bioEitherFn: BIO3[EitherFn] = null
    try threeTo2[EitherFn]
    catch { case _: NullPointerException => }
```

VladPodilnyk: also, such test  passes
```scala
assertCompiles(
      """import izumi.functional.bio.{F, BIOMonad, BIOMonadAsk, BIOPrimitives, BIORef3}
        |
        |def adder[F[+_, +_]: BIOMonad: BIOPrimitives](i: Int): F[Nothing, Int] =
        |  F.mkRef(0)
        |   .flatMap(ref => ref.update(_ + i) *> ref.get)
        |""".stripMargin)
```

VladPodilnyk: this assert passes
```scala
assertCompiles("""import izumi.functional.bio.{F, BIOMonad, BIOMonadAsk, BIOPrimitives, BIORef3}
                     |
                     |def adder[F[+_, +_]: BIOMonad: BIOPrimitives](i: Int): F[Nothing, Int] =
                     |  F.mkRef(0)
                     |   .flatMap(ref => ref.update(_ + i) *> ref.get)
                     |""".stripMargin)
```

neko-kai (approved):

PR #1011, opened 2020-04-18
distage-testkit: Fix recorded test error messages in sbt reports, scalatest xml reports, etc. 

neko-kai:
@alexander-branevskiy 

* preserve original class of Throwable in BIORunner.unsafeRun
* add BIOExit.Trace.unsafeAttachTrace


alexander-branevskiy:
LGTM

PR #1016, opened 2020-04-21
Fix scanner

alexander-branevskiy:


alexander-branevskiy:
i will open another one with a better history

PR #1017, opened 2020-04-21
fix scanner

alexander-branevskiy:


neko-kai:
@alexander-branevskiy Thank you! 🙏

PR #1029, opened 2020-05-03
Added zipWithPar and derived zipPar, zipParLeft and zipParRight

### LINKED ISSUE #1028: Add parallel zip to BIOAsync3
`BIOAsync3` currently allows for `parTraverse` and `race` but it is missing zipping in parallel.

Adriani277:
closes #1028

### DIFF
--- a/fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/BIOAsync3.scala
+++ b/fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/BIOAsync3.scala
@@ -23,6 +23,13 @@ trait BIOAsync3[F[-_, +_, +_]] extends BIO3[F] {
   def parTraverseN[R, E, A, B](maxConcurrent: Int)(l: Iterable[A])(f: A => F[R, E, B]): F[R, E, List[B]]
   def parTraverse[R, E, A, B](l: Iterable[A])(f: A => F[R, E, B]): F[R, E, List[B]]
 
+  /**
+    * Returns an effect that executes both effects,
+    * in parallel, combining their results with the specified `f` function. If
+    * either side fails, then the other side will be interrupted.
+    */
+  def zipWithPar[R, E, A, R1 <: R, E1 >: E, B, C](fa: F[R, E, A], fb: F[R1, E1, B])(f: (A, B) => C): F[R1, E1, C]
+
   def uninterruptible[R, E, A](r: F[R, E, A]): F[R, E, A]
 
   // defaults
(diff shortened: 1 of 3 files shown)


Adriani277:
@neko-kai I know you have mentioned possibly adding a `BIOParallel` but I was not sure where this would fit in the hierarchy. 
I assume `BIOParallel` would be extended by `BIOAsync`?

I have 2 commits:

1. Adding functions to `BIOAsync3`
2. An attempt at creating a `BIOParallel` sub-hierarchy

Please let me know your thoughts on this

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/BIOAsync3.scala
```diff
@@ -31,5 +38,26 @@ trait BIOAsync3[F[-_, +_, +_]] extends BIO3[F] {
   def parTraverse_[R, E, A, B](l: Iterable[A])(f: A => F[R, E, B]): F[R, E, Unit] = void(parTraverse(l)(f))
   def parTraverseN_[R, E, A, B](maxConcurrent: Int)(l: Iterable[A])(f: A => F[R, E, B]): F[R, E, Unit] = void(parTraverseN(maxConcurrent)(l)(f))
 
+  /**
+    * Returns an effect that executes both effects,
+    * in parallel, combining their results into a tuple. If either side fails,
+    * then the other side will be interrupted.
+    */
+  final def zipPar[R, E, A, R1 <: R, E1 >: E, B](fa: F[R, E, A], fb: F[R1, E1, B]): F[R1, E1, (A, B)] = zipWithPar(fa, fb)((a, b) => (a, b))
```
Adriani277, line 46: I have noticed that `parTraverseN_` is not `final`. Should this also not be `final`?

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/BIOParallel.scala
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+    * Returns an effect that executes both effects,
+    * in parallel, combining their results with the specified `f` function. If
+    * either side fails, then the other side will be interrupted.
+    */
+  def zipWithPar[R, E, A, R1 <: R, E1 >: E, B, C](fa: F[R, E, A], fb: F[R1, E1, B])(f: (A, B) => C): F[R1, E1, C]
+
+  // defaults
+  def parTraverse_[R, E, A, B](l: Iterable[A])(f: A => F[R, E, B]): F[R, E, Unit] = InnerF.void(parTraverse(l)(f))
```
Adriani277, line 17: I have noticed that `parTraverse_` is not `final`. Should this be made `final`? If not, why?

neko-kai: @Adriani277
This shouldn't be made `final` because `BIOAsyncZio` instance overrides it to use `ZIO.foreachPar_`. This lets it use a native implementation from the effect type that is more optimized and may have enhancements compared to a naive implementation (like, better support for stack-traces, e.g. notice the `new ZipLeftFn` wrapper in implementation of `ZIO.<*` - https://github.com/zio/zio/blob/master/core/shared/src/main/scala/zio/ZIO.scala#L108 - this allows the stacktrace where `<*` occurs to point to the source code line contained inside the user-provided `that` lambda, instead of the line in ZIO.scala source file that would be there without the `new ZipLeftFn`)

Generally, the only methods that are final in the hierarchy are those that are   trivial enough they _can't_ have better native implementations in the effect type. So, I'd propose to un-final `zipParLeft` & `zipParRight` and redirect them to `<&` and `&>` in `BIOAsyncZio`

neko-kai:
@Adriani277 
I imagined `BIOParallel` would be entirely off-side, like `cats.Parallel`, but actually I like what you did much better, it makes a lot more sense, thanks!

Also, the `BIOParallel` class should be enough to provide a conversion to `cats.Parallel` in `BIOCatsConversions` - https://github.com/7mind/izumi/blob/111390eab94e03be343913fe8167a7e048ad24a9/fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/BIOCatsConversions.scala#L55 - if you're interested, you could add it in a subsequent PR

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/BIOParallel.scala
Adriani277: I see, thanks for the explanation.

Will make the changes

Adriani277:
I can have a look at adding the parallel conversion next.
I would also like to add `BIO3Syntax` for `BIOParallel` but I don't understand it enough yet.
Will attempt in the subsequent PR, unless you think `BIOParallel` should not have `syntax` added

neko-kai (approved):

neko-kai:
@Adriani277 
Adding syntax to `BIOParallel` would involve

* adding a superclass to `BIOAsyncOps` here - https://github.com/7mind/izumi/blob/111390eab94e03be343913fe8167a7e048ad24a9/fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/syntax/BIOSyntax.scala#L135
* Adding overloads named `BIOParallel` to `BIOImplicitPuns2` (check the `BIOAsync overloads above for example) - https://github.com/7mind/izumi/blob/111390eab94e03be343913fe8167a7e048ad24a9/fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/syntax/BIOSyntax.scala#L164
* Shifting all the ImplicitPuns starting from `BIO` down by one, such that `BIOImplicitPuns11` is now `BIOImplicitPuns12` https://github.com/7mind/izumi/blob/111390eab94e03be343913fe8167a7e048ad24a9/fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/syntax/BIOSyntax.scala#L209

This is for bifunctor BIOSyntax, you'll also need to repeat the same steps for trifunctor `BIOSyntax3`.

You'll also need to rename `BIOParallel` to `BIOParallel3` and add a type alias `BIOParalell` into `bio` package object here - https://github.com/7mind/izumi/blob/111390eab94e03be343913fe8167a7e048ad24a9/fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/package.scala#L74

Adriani277:
@neko-kai awesome. Will have a go at adding it

PR #845, opened 2020-01-21
0.11.0: Mutators

neko-kai:

PR #1046, opened 2020-05-07
Global Planning Policy

pshirshov:
Completely formal ahead of time conflict resolution. A necessary step to introduce mutators.


neko-kai:
Moved to https://github.com/7mind/izumi/pull/845

PR #1064, opened 2020-05-18
remove inter-test dependencies on distage-core/test module to build more modules in parallel

neko-kai:


neko-kai:
not clear yet if there's any (positive) impact

PR #1072, opened 2020-05-22
distage-testkit: Smart memoization group merging.

### LINKED ISSUE #1067: distage-testkit: Smart memoization group merging
Currently, *any* difference in *any* field of `TestConfig` will result in creation of a new memoization group and a repeat allocation of all Global, memoized components within that group. 

We must have different memoization groups because memoized components can be affected by their dependencies, by axis etc. e.g. if we memoize `PetsRepo[F]`, we want to have a different `PetsRepo` for `Repo.Dummy` & `Repo.Prod` activation axes - we can't share the implementation between them.

However, the current strategy is too simplistic and extemely coarse grained because ANY change creates a new env, you are forced to not use such things as `TestConfig#moduleOverrides` if you don't want to lose memoization. E.g. if we have a component `LlamaShaver[F]` with implementations in axises `Shaver.Dull` & `Shaver.Sharp` – changing that axis should have NO EFFECT on the ability to share `PetsRepo[F]`, because the component affected by the change is not a dependency of PetsRepo and does not affect the shared Locator.

Plan: First try a totally-straightforward approach, just prepare memoization plans via `trisectByKeys` for all the different `TestConfig`s same as currently - and then compare the resulting plans using `equals` = if they are the same then instantiate the plan once and reuse the Locator. If it works this would allow merging test envs entirely post-hoc without any complex rules. It _can_ work fine because `Planner` is pure and all components of an `OrderedPlan`, including `ProviderMagnet` have fairly reasonable `equals`.

Caparow:
fix #1067

### DIFF
--- a/distage/distage-testkit-core/src/main/scala/izumi/distage/testkit/services/dstest/DistageTestRunner.scala
+++ b/distage/distage-testkit-core/src/main/scala/izumi/distage/testkit/services/dstest/DistageTestRunner.scala
@@ -38,6 +38,7 @@ class DistageTestRunner[F[_]: TagK](
 ) {
   def run(): Unit = {
     val envs = groupTests(tests)
+    logEnvironmentsInfo(envs)
     try {
       val (parallelEnvs, sequentialEnvs) = envs.partition(_._1.env.parallelEnvs)
       runEnvs(parallel = true)(parallelEnvs)
@@ -63,8 +64,7 @@ class DistageTestRunner[F[_]: TagK](
     // here we are grouping our tests by memoization env
     distageTests.groupBy(_.environment.toMemoizationEnv).flatMap {
       case (memoEnv, grouped) =>
-        val memoizedEnvLogger =
-          IzLogger(getTestRunnerLogLevel(memoEnv.verboseTestRunner)).withCustomContext(testRunnerLogger.customContext)("memoEnv" -> memoEnv.hashCode())
+        val memoizedEnvLogger = IzLogger(getTestRunnerLogLevel(memoEnv.verboseTestRunner)).withCustomContext(testRunnerLogger.customContext)
         grouped
           .groupBy(_.environment).map {
             case (env, tests) =>
@@ -100,7 +100,7 @@ class DistageTestRunner[F[_]: TagK](
                   acc -> b
               }
               if (strengthenedKeys.nonEmpty) {
-                memoizedEnvLogger.debug(s"Strengthened weak components: $strengthenedKeys")
+                memoizedEnvLogger.debug(s"Strengthened weak components in ${memoEnv.hashCode() -> "memoEnv"}: $strengthenedKeys")
               }
 
               // compute [[TriSplittedPlan]] of our test, to extract shared plan, and perform it only once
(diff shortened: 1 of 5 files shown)


distage/distage-testkit-core/src/main/scala/izumi/distage/testkit/services/dstest/DistageTestRunner.scala
```diff
@@ -60,146 +77,208 @@ class DistageTestRunner[F[_]: TagK]
...
+              use(integrationSharedLocator)
+            }
         }
+    }
+  }
 
-        val shared = injector.trisectByKeys(strengthenedAppModule, sharedKeys) {
-          _.collectChildren[IntegrationCheck].map(_.target).toSet
-        }
+  def withTestsRecoverCase(tests: => Iterable[DistageTest[F]])(testsAction: F[Unit])(implicit F: DIEffect[F]): F[Unit] = {
```
neko-kai: `testsAction` should be by-name because `F` is not really a safe/sane effect type, it can be `Identity`

neko-kai:
A thought: we can limit the amount of logs for re-reading the same config files by memoizing config reads with the same `testBaseName` (for example using a mutable map and `getOrElseUpdate`)

Caparow:
> A thought: we can limit the amount of logs for re-reading the same config files by memoizing config reads with the same `testBaseName` (for example using a mutable map and `getOrElseUpdate`)

A good idea!
Will do

neko-kai:
Another thought, we can replace `testRunnerLogLevel` with just a switch `debugOutput: Boolean` and set level `Debug` on `true`, `Info` on `false`, and _force_ level to `Debug` if property `DebugProperties.\`izumi.distage.testkit.debug\`` is set to true.
Then we can also remove the separate `debugLogger` and just log with `IzLogger.debug`.

Caparow:
> Another thought, we can replace `testRunnerLogLevel` with just a switch `debugOutput: Boolean` and set level `Debug` on `true`, `Info` on `false`, and _force_ level to `Debug` if property `DebugProperties.\`izumi.distage.testkit.debug``is set to true. Then we can also remove the separate`debugLogger`and just log with`IzLogger.debug`.

But by this implementation we will lose control over the test log level. It's okay to have `info` level always, but sometimes we need to force `warn` level to get rid of the noise.

neko-kai:
@Caparow Hmm, actually `earlyLogger` is used just once - when loading config, and currently there's no way to disable config loader printing - afterwards logger settings are read from config. Hmm, we may alter the log levels of the debug messages themselves (promote Debug to Info) instead of chaging the logger. We may also have three values for `debugOutput` = `Silent/Normal/Verbose`

Caparow:
> @Caparow Hmm, actually `earlyLogger` is used just once - when loading config, and currently there's no way to disable config loader printing - afterwards logger settings are read from config. Hmm, we may alter the log levels of the debug messages themselves (promote Debug to Info) instead of chaging the logger. We may also have three values for `debugOutput` = `Silent/Normal/Verbose`

You still can silence the config loader. Early logger uses `testRunnerLogLevel`. Also, a late logger is inherited from early logger. And there is no point to add new log level ADT (at least I don't know why we'll need this).

neko-kai (approved):

PR #1074, opened 2020-05-23
distage-testkit: Add `TestConfig#debugOutput`, rename `TestConfig#testRunnerLogLevel`->`logLevel`, choose lowest `logLevel` among all tests envs for test runner logs (allow silencing config load log).

neko-kai:
* system property `izumi.distage.testkit.debug` sets testkit debug logs to info, instead of setting logger to Debug level
* log how many envs were merged if any
* exclude _all_ components of runtimeLocator from subsequent plans
* fixed memoizedConfig for multiple test effect types
* update TestConfig doc

### DIFF
--- a/distage/distage-testkit-core/src/main/scala/izumi/distage/testkit/services/dstest/DistageTestRunner.scala
+++ b/distage/distage-testkit-core/src/main/scala/izumi/distage/testkit/services/dstest/DistageTestRunner.scala
@@ -60,77 +60,89 @@ class DistageTestRunner[F[_]: TagK](
     *  By result you'll got [[MemoizationEnvWithPlan]] mapped to tests wit it's plans.
     *  [[MemoizationEnvWithPlan]] represents memoization environment, with shared [[TriSplittedPlan]], [[Injector]], and runtime plan.
     * */
-  def groupTests(distageTests: Seq[DistageTest[F]]): Iterable[(MemoizationEnvWithPlan, Iterable[(DistageTest[F], OrderedPlan)])] = {
+  def groupTests(distageTests: Seq[DistageTest[F]]): Map[MemoizationEnvWithPlan, Seq[(DistageTest[F], OrderedPlan)]] = {
+
+    final case class EnvByproducts(injector: Injector, envLogger: IzLogger)
+
     // here we are grouping our tests by memoization env
     distageTests.groupBy(_.environment.toMemoizationEnv).flatMap {
       case (memoEnv, grouped) =>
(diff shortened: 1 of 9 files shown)


distage/distage-testkit-core/src/main/scala/izumi/distage/testkit/services/dstest/DistageTestRunner.scala
```diff
@@ -60,77 +60,89 @@ class DistageTestRunner[F[_]: TagK](
...
-        val memoizedEnvLogger = IzLogger(getTestRunnerLogLevel(memoEnv.verboseTestRunner)).withCustomContext(testRunnerLogger.customContext)
-        grouped
-          .groupBy(_.environment).map {
-            case (env, tests) =>
-              // make a config loader for current env with logger
-              val config = loadConfig(env, memoizedEnvLogger)
-              val runtimeLogger = EarlyLoggers.makeLateLogger(RawAppArgs.empty, memoizedEnvLogger, config, env.logLevel, defaultLogFormatJson = false)(
-                "memoEnv" -> memoEnv.hashCode()
-              )
+        val configLoadLogger = IzLogger(memoEnv.logLevel).withCustomContext("phase" -> "testRunner")
```
Caparow, line 70: Hm, but with such behavior you cannot control config logger log level via debug property.

neko-kai: @Caparow Yeah, because it's on by default - you should bump `logLevel` to filter it out. I could add a debug property to _disable_ config logs, but i think it wouldn't be that useful - it's kinda hard to add persistent system properties everywhere, if you're the kind of person who wants zero-log testing

neko-kai: Or we can use an enum instead of bool after all - then you could disable config log by setting `Silent`

Caparow: @neko-kai ah, I just thought that we want to make config loader with test runner logger instead of environment one (because making a config is a part of env preparation rather than test run). Anyway, it's not necessary since we already reduced the number of logs for config loader, and in the vast majority of cases we will have only one config loader message.

neko-kai: Ok in that case let's merge as-is

PR #1078, opened 2020-05-24
distage-testkit: add config for parallelism level

### LINKED ISSUE #1047: distage-testkit: add config to limit parallelism level
We have fields in `TestConfig` to enable/disable parallelism, but we have no config to _limit_ the parallelism level. It's _implied_ that the thread-pool size implicitly limits the parallelism, but it may not always be the case.

Caparow:
Fix #1047

### DIFF
--- a/distage/distage-core-api/src/main/scala/izumi/distage/model/effect/DIEffectAsync.scala
+++ b/distage/distage-core-api/src/main/scala/izumi/distage/model/effect/DIEffectAsync.scala
@@ -3,9 +3,9 @@ package izumi.distage.model.effect
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.concurrent.{Executors, ThreadFactory}
 
-import cats.Parallel
-import cats.effect.Timer
-import izumi.distage.model.effect.LowPriorityDIEffectAsyncInstances.{_Parallel, _Timer}
+import cats.{Functor, Parallel}
+import cats.effect.{Concurrent, Timer}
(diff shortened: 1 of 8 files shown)


distage/distage-core-api/src/main/scala/izumi/distage/model/effect/DIEffectAsync.scala
```diff
@@ -93,7 +112,7 @@ private[effect] sealed trait LowPriorityDIEffectAsyncInstances {
     *
     * Optional instance via https://blog.7mind.io/no-more-orphans.html
     */
-  implicit final def fromParallelTimer[F[_], P[_[_]]: _Parallel, T[_[_]]: _Timer](implicit P: P[F], T: T[F]): DIEffectAsync[F] = {
+  implicit final def fromCats[F[_], P[_[_]]: _Parallel, T[_[_]]: _Timer, C[_[_]]: _Concurrent, M[_[_]]: _Functor](implicit P: P[F], T: T[F], C: C[F], M: M[F]): DIEffectAsync[F] = {
```
neko-kai: Concurrent extends Functor, so additional Functor is redundant

neko-kai (approved):

distage/distage-core-api/src/main/scala/izumi/distage/model/effect/DIEffectAsync.scala
Caparow: Yup, fixed it already.

PR #1085, opened 2020-05-28
distage-framework-docker: add defaults to Docker.ClientConfig constructor

coreyoconnor:
Defaults are from `docker-reference.conf` except for `remote` and `registry`. I think it would be odd to have default parameters that are examples values (eg: `dockeruser`). Necessary for the `reference.conf` to show structure, but seems out of place here.


neko-kai (approved):
Thank you! :+1:

PR #1088, opened 2020-05-29
Feature/add replace parameter ids on binding for mutator syntax

### LINKED ISSUE #444: Add/replace parameter ids on binding level
We have the following common case

Library code:

```
class LibClass(context: ExecutionContext)
```

client code:
```
make[LibClass].from {
  ec: ExecutionContext @Id("blocking")  =>
    new LibClass(ec)
}
```

it would be nice to have better syntax:

```
make[LibClass].annotateParameter[Dependency1]("value")
```

Caparow:
Closes #444

### DIFF
--- a/distage/distage-core-api/src/main/scala/izumi/distage/model/definition/dsl/AbstractBindingDefDSL.scala
+++ b/distage/distage-core-api/src/main/scala/izumi/distage/model/definition/dsl/AbstractBindingDefDSL.scala
@@ -161,45 +161,42 @@ object AbstractBindingDefDSL {
     override def interpret: collection.Seq[ImplBinding] = {
       var b: SingletonBinding[DIKey.BasicKey] = initial
       var refs: List[SingletonBinding[DIKey.BasicKey]] = Nil
-
-      println(ops)
-
-      ops
-        .sortBy {
-          case _: SetImpl => 0
-          case _: AddTags => 0
-          case _: SetParameterId[_] => 1
-          case _: SetId[_] => 0
-          case _: SetIdFromImplName => 0
-          case _: AliasTo => 0
-        }.foreach {
-          case SetImpl(implDef) =>
-            b = b.withImplDef(implDef)
-          case AddTags(tags) =>
-            b = b.addTags(tags)
-          case SetId(id, idContract) =>
-            val key = DIKey.IdKey(b.key.tpe, id)(idContract)
-            b = b.withTarget(key)
-          case SetIdFromImplName() =>
-            // b.key.tpe is the same b.implementation.tpe because `SetIdFromImplName` comes before `SetImpl`...
-            b = b.withTarget(DIKey.IdKey(b.key.tpe, b.key.tpe.tag.longName.toString.toLowerCase))
-          case AliasTo(key, pos) =>
-            // it's ok to retrieve `tags`, `implType` & `key` from `b` because all changes to
-            // `b` properties must come before first `aliased` call
-            // after first `aliased` no more changes are possible
-            val newRef = SingletonBinding(key, ImplDef.ReferenceImpl(b.implementation.implType, b.key, weak = false), b.tags, pos)
-            refs = newRef :: refs
-          case SetParameterId(key, id, idContract) =>
-            b.implementation match {
-              case ImplDef.ProviderImpl(implType, function) =>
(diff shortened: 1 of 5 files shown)


distage/distage-core-api/src/main/scala/izumi/distage/model/definition/dsl/ModuleDefDSL.scala
```diff
@@ -745,31 +768,17 @@ object ModuleDefDSL {
...
-    override protected val mutableState: SingletonRef,
-    override protected val key: DIKey.IdKey[_],
-  ) extends MakeDSLBase[T, MakeDSLNamedAfterFrom[T]]
-    with MakeDSLMutBase[T] {
-
-    def tagged(tags: BindingTag*): MakeNamedDSL[T] = {
-      addOp(AddTags(tags.toSet)) {
-        new MakeNamedDSL[T](_, key)
-      }
+    def annotateParameter[P]: PartiallyAppliedParameterAnnotation[P, MakeDSLUnnamedAfterFrom[T]] = {
```
neko-kai, line 775: need to add id to `MakeNamedDSL` & `MakeDSLNamedAfterFrom`

distage/distage-core-api/src/main/scala/izumi/distage/model/definition/dsl/AbstractBindingDefDSL.scala
```diff
@@ -200,9 +207,15 @@ object AbstractBindingDefDSL {
...
         case Modify(providerMagnetModifier) =>
-          ???
-//          val newRef = SingletonBinding(key, ImplDef.ProviderImpl(b.implementation.implType, ???), b.tags, ???)
-//          refs = newRef :: refs
+          b.implementation match {
+            case ImplDef.ProviderImpl(implType, function) =>
+              val oldMagnet = ProviderMagnet(function)
+              val newProvider = providerMagnetModifier(oldMagnet).get
+              if (implType == newProvider.ret) {
+                b = b.withImplDef(ImplDef.ProviderImpl(newProvider.ret, newProvider))
```
neko-kai: Why conditional update? We forbid incompatible types at API level & it's safe for the new type to be `<:` than less previous `implType`

distage/distage-core/src/test/scala/izumi/distage/dsl/DSLTest.scala
```diff
@@ -67,6 +67,27 @@ class DSLTest extends AnyWordSpec with MkInjector {
       )
     }
 
+    "support annotated parameters after from" in {
```
neko-kai: Could you please also add a test that uses the injector to instantiate such a module?

distage/distage-core-api/src/main/scala/izumi/distage/model/definition/dsl/ModuleDefDSL.scala
Caparow: Done.

distage/distage-core/src/test/scala/izumi/distage/dsl/DSLTest.scala
Caparow: Done.

distage/distage-core-api/src/main/scala/izumi/distage/model/definition/dsl/AbstractBindingDefDSL.scala
Caparow: Yeah, you are right. I've added this check instead: `if (newProvider.ret <:< implType)`. Is this enough?

distage/distage-core-api/src/main/scala/izumi/distage/model/definition/dsl/AbstractBindingDefDSL.scala
```diff
@@ -211,8 +211,8 @@ object AbstractBindingDefDSL {
             case ImplDef.ProviderImpl(implType, function) =>
               val oldMagnet = ProviderMagnet(function)
               val newProvider = providerMagnetModifier(oldMagnet).get
-              if (implType == newProvider.ret) {
-                b = b.withImplDef(ImplDef.ProviderImpl(newProvider.ret, newProvider))
+              if (newProvider.ret <:< implType) {
```
neko-kai, line 194: @Caparow  Are you sure the condition is necessary at all, since we already restrict types? If so we should at least have an `else` branch throwing an exception, so that no silent ignoring happens if the condition doesn't apply.

neko-kai (approved):

PR #1104, opened 2020-06-08
distage-docker: reuse exited containers

### LINKED ISSUE #1042: Docker testkit will try to restart non-daemon containers that stopped themselves, even if `reuse` flag is enabled
The current implementation of `ContainerResource` will check only running containers for determining should we run a new container or was it run before. This leads to multiple re-runnings of containers that will terminate after execution, which can affect test execution time. 

**E.g.:** 
If we will start Postgres container, and dependent on that Flyway (to migrate DB) with `reuse` flag on both containers - then Postgres container will be actually reused, but Flyway container will be starting for each test environment. (As Flyway container will be stopped each time after successful migration)

**Possible solutions:**

 - Add a new config option like `reuse` and lock on container start for a whole test run scope, to run only one such container in time. On the start of tests with such containers, we will create a temp file with the name of the container, and during next test environment start we will check this file on existence (if file exists - the container was started and we can "reuse" this, otherways - start a container and acquire the lock). But such flow will start containers with this option on each test run.

 - Add a new config option like `reuse` and use `docker ps -a` to find what containers were run before. But this method will actually need `docker container prune` execution to drop all unused containers, because the current containers list may contain old ones. Without cleaning (or filtering) containers we could actually skip container startup what even worse than re-running. But container prune also will drop all user-defined containers. Also, this method just like previous will re-start containers on each test run if we will use pruning (or filtering). 

- Combine both solutions: tag each container with some value, which would be unique for one test run and during container start - try to find container that belongs to the current run.

Caparow:
Closes #1042

### DIFF
--- a/distage/distage-framework-docker/src/main/resources/docker-reference.conf
+++ b/distage/distage-framework-docker/src/main/resources/docker-reference.conf
@@ -1,5 +1,5 @@
 docker {
-  readTimeoutMs = 5000
+  readTimeoutMs = 60000
   connectTimeoutMs = 1000
   allowReuse = true
 
--- a/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/ContainerNetworkDef.scala
+++ b/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/ContainerNetworkDef.scala
@@ -36,11 +36,11 @@ object ContainerNetworkDef {
 
   final class NetworkResource[F[_]: DIEffect: DIEffectAsync, T](
     config: ContainerNetworkConfig[T],
-    clientw: DockerClientWrapper[F],
+    client: DockerClientWrapper[F],
     prefixName: String,
     logger: IzLogger,
   ) extends DIResource[F, ContainerNetwork[T]] {
-    private[this] val client = clientw.rawClient
+    private[this] val rawClient = client.rawClient
     private[this] val prefix = prefixName.camelToUnderscores.drop(1).replace("$", "")
     private[this] val stableLabels: Map[String, String] = Map(
       "distage.reuse" -> shouldReuse(config),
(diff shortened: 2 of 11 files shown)


neko-kai:
@Caparow 
Getting compile errors on CI currently:
```
[error] /home/vsts/work/1/s/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/ContainerResource.scala:143:37: value when is not a member of object Option
[error]           val statusFilter = Option.when(ports.isEmpty)("exited").toList ++ List("running")
[error]                                     ^
[error] /home/vsts/work/1/s/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/ContainerResource.scala:190:34: value getId is not a member of Any
[error]               id = ContainerId(c.getId),
[error]                                  ^
[error] /home/vsts/work/1/s/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/ContainerResource.scala:191:33: value getName is not a member of Any
[error]               name = inspection.getName,
[error]                                 ^
[error] /home/vsts/work/1/s/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/ContainerResource.scala:196:37: value getConfig is not a member of Any
[error]               hostName = inspection.getConfig.getHostName,
[error]                                     ^
[error] /home/vsts/work/1/s/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/ContainerResource.scala:197:35: value getConfig is not a member of Any
[error]               labels = inspection.getConfig.getLabels.asScala.toMap,
[error]                                   ^
[error] /home/vsts/work/1/s/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/ContainerResource.scala:192:30: type mismatch;
[error]  found   : Any
[error]  required: izumi.distage.docker.Docker.ContainerConnectivity
[error]               connectivity = existingPorts,
[error]                              ^
[error] 6 errors found
[info] Compiling 29 Scala sources to /home/vsts/work
```

neko-kai (approved):

PR #1117, opened 2020-06-18
Add async methods to DIEffectAsync, use async file lock for file mutex.

### LINKED ISSUE #1111: distage-framework-docker: use AsynchronousFileChannel instead of blocking wait loops on file locks
This will require adding `async` method to `DIEffectAsync`
```scala
def async[A](effect: (Either[Throwable, A] => Unit) => Unit): F[A]
```
but will allow removing repeat loops in favor of just waiting on the timeout asynchronously.

For `Identity` `async` will have to be implemented with `Await.result` on `scala.concurrent.Promise`

/cc @Caparow

Caparow:
Closes #1111

### DIFF
--- a/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/FileLockMutex.scala
+++ b/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/FileLockMutex.scala
@@ -22,7 +22,7 @@ object FileLockMutex {
     F: DIEffect[F],
     P: DIEffectAsync[F],
   ): F[E] = {
-    def retryOnFileLock[A](eff: F[Option[A]], attempts: Int = 0): F[Option[A]] = {
+    def retryOnFileLock(eff: F[Option[FileLock]], attempts: Int = 0): F[Option[FileLock]] = {
       logger.debug(s"Attempt ${attempts -> "num"} out of $maxAttempts to acquire lock for $filename.")
       F.definitelyRecover(eff) {
         case _: OverlappingFileLockException =>
--- a/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/FileLockMutex.scala
+++ b/distage/distage-framework-docker/src/main/scala/izumi/distage/docker/FileLockMutex.scala
@@ -22,9 +22,9 @@ object FileLockMutex {
     F: DIEffect[F],
     P: DIEffectAsync[F],
   ): F[E] = {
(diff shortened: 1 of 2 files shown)


distage/distage-framework-docker/src/main/scala/izumi/distage/docker/FileLockMutex.scala
```diff
@@ -22,23 +22,18 @@ object FileLockMutex {
...
-          } catch {
-            case _: OverlappingFileLockException => None
-          }
-        }.flatMap {
-          case Some(lock) =>
-            effect.guarantee(F.maybeSuspend(lock.close()))
-          case None if attempts < maxAttempts =>
-            P.sleep(waitFor).flatMap(_ => acquireAndRun(channel, attempts + 1))
-          case None =>
+    def retryOnFileLock(eff: F[Option[FileLock]], attempts: Int = 0): F[Option[FileLock]] = {
```
neko-kai:
```suggestion
    def retryOnFileLock(eff: F[FileLock], attempts: Int = 0): F[Option[FileLock]] = {
```
might wrap into Some later since `eff` cannot return None now

distage/distage-framework-docker/src/main/scala/izumi/distage/docker/FileLockMutex.scala
```diff
@@ -47,10 +42,23 @@ object FileLockMutex {
...
+        retryOnFileLock {
+          P.async[Option[FileLock]] {
+            cb =>
+              val handler = new CompletionHandler[FileLock, Unit] {
+                override def completed(result: FileLock, attachment: Unit): Unit = cb(Right(Some(result)))
+                override def failed(exc: Throwable, attachment: Unit): Unit = cb(Left(exc))
+              }
+              channel.lock((), handler)
+          }
+        }.flatMap {
```
neko-kai: This looks like it should use `bracket` instead

Caparow: Done.

distage/distage-framework-docker/src/main/scala/izumi/distage/docker/FileLockMutex.scala
Caparow: Done.

neko-kai (approved):

PR #1153, opened 2020-07-23
Add activation axis to forced and memoization roots.

### LINKED ISSUE #1044: distage-testkit: Add axes to `forcedRoots` & `memoizationRoots` in TestConfig
Adding axes to the DI roots will allow us to filter roots in depending on chosen axis.

E.g.
```scala
abstract class FooTest extends DistageBIOSpecScalatest[IO] {
   override protected def config: TestConfig = super.config(
      forcedRoots = Map(Set(Repo.Prod) -> Set(DIKey.get[PostgresDDL]))
   )
}

// `PostgresDDL` would be filtered from the DI roots
final class DummyRepoTest extends FooTest {
   override protected def config: TestConfig = super.config(
      activation = Activation(Repo -> Repo.Dummy, ExternalApi -> ExternalApi.Mock)
   )
  ...
}

// `PostgresDDL` would be persist in DI roots and will perform tables ddl up
final class ProductionRepoTest extends FooTest {
   override protected def config: TestConfig = super.config(
      activation = Activation(Repo -> Repo.Prod, ExternalApi -> ExternalApi.Prod)
   )
  ...
}
```

Caparow:
Closes #1044


pshirshov (commented):
I'm not sure why do we need this. Axis are intended to be user configurable, while there is no sane scenario when user may need to alter cycle resolution policy through application commandline.

pshirshov (approved):

Issue #1001, opened 2020-04-14
[LogStage] Specify fields in JSON


gu1p:
Hey, guys! Congratulations on the job! :)

So, is there a way to specify the exact shape of each JSON line produced (eliminating fields like "template", "thread", "meta" etc)?

neko-kai:
You need to supply a different `RenderingPolicy` for a `LogSink`, e.g. by overriding `LogstageCirceRenderingPolicy`, so instead of:

```scala
IzLogger(Log.Level.Info, Seq(ConsoleSink(LogstageCirceRenderingPolicy())))
```

instantiate with e.g.

```scala
IzLogger(Log.Level.Info, Seq(ConsoleSink(new LogstageCirceRenderingPolicy() {
  override protected def makeEventEnvelope(entry: Log.Entry, formatted: RenderedMessage): Seq[(String, Json)] = ...
})))
```

gu1p:
Thanks @neko-kai!

tuleism:
@neko-kai 

What's your thought on a PR to make JSON key easily overridable. For example, instead of:

```scala
if (ctx.nonEmpty) {
  result += "context" -> ctx.asJson
}
```

we have

```scala
protected val ContextKey = "context"
if (ctx.nonEmpty) {
  result += ContextKey -> ctx.asJson
}
```

neko-kai:
@tuleism 
Sure, that would be fine

pshirshov:
I guess I'll close this as per we are preparing 0.11 release at the moment and the fix is accepted.

PR #1167, opened 2020-08-02
logstage-rendering-circe: make JSON key overridable

tuleism:
Discussion at https://github.com/7mind/izumi/issues/1001#issuecomment-660983314

### DIFF
--- a/logstage/logstage-rendering-circe/src/main/scala/izumi/logstage/api/rendering/json/LogstageCirceRenderingPolicy.scala
+++ b/logstage/logstage-rendering-circe/src/main/scala/izumi/logstage/api/rendering/json/LogstageCirceRenderingPolicy.scala
@@ -17,6 +17,11 @@ class LogstageCirceRenderingPolicy(
 
   import LogstageCirceRenderingPolicy._
 
+  protected def EventKey = "event"
+  protected def ContextKey = "context"
+  protected def MetaKey = "meta"
+  protected def TextKey = "text"
+
   override def render(entry: Log.Entry): String = {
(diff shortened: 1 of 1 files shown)


logstage/logstage-rendering-circe/src/main/scala/izumi/logstage/api/rendering/json/LogstageCirceRenderingPolicy.scala
```diff
@@ -17,6 +17,11 @@ class LogstageCirceRenderingPolicy(
 
   import LogstageCirceRenderingPolicy._
 
+  protected val EventKey = "event"
+  protected val ContextKey = "context"
+  protected val MetaKey = "meta"
+  protected val TextKey = "text"
```
neko-kai:
```suggestion
  protected def textKey = "text"
```
This should probably be a `def`, not `val`, because defs for constants are actually more efficient (and easier to override than vals) – constants are stored in the constant section in the classfile and subsequently in the constant pool on the JVM heap, whenever a method returns a constant it will return an address from the constant pool - it will not allocate a new string. Now a `val` also makes a field in the class that stores the address of the constant, but this is superfluous since a method would already return a pointer to a string from the constant pool, a val would just inflate the in-memory size of the object with a copy of that pointer inside the _object_ (that is already stored in the _class_ after it's loaded)

tuleism: Fixed. Thanks for your detail explanation!

neko-kai (approved):
Thank you for your contribution! 🙏

PR #1170, opened 2020-08-04
distage-framework-docker: Postgres with FlyWay container.

### LINKED ISSUE #1158: distage-framework-docker: Add Flyway container
There were some questions about how to perform migrations on test containers, we may answer them in code with a ready flyway container

Caparow:
Closes #1158


distage/distage-framework-docker/src/main/resources/sql/V1__Baseline.sql
```diff
@@ -0,0 +1 @@
+select now();
```
neko-kai: I guess this should be in `test/` resources not main?

distage/distage-framework-docker/src/main/scala/izumi/distage/docker/bundled/PostgresFlyWayDocker.scala
```diff
@@ -0,0 +1,80 @@
...
+      cmd = Seq(
+        "-url=jdbc:postgresql://localhost/postgres",
+        "-schemas=public",
+        "-user=postgres",
+        "-password=postgres",
+        "-connectRetries=60",
+        "baseline",
+        "migrate",
+      ),
+      mounts = Seq(Mount(this.getClass.getClassLoader.getResource("sql").getPath, "/flyway/sql")),
```
neko-kai: This will not work when resources are in jars (e.g. outside of dev env)
I think we need to allow parameterization to allow user to choose a suiteable location to bind...

Caparow: Yeah, I was thinking about the creation of a container with the configuration for this purpose. But decided to implement a hardcoded version to unify with the other containers style (which are objects) and wait till review. Because current independent Postgres container implementation also contains hardcoded values and hardly can be used in real-world tests.

Caparow: I'll update this container then.

Caparow: @neko-kai Done.

neko-kai (approved):

neko-kai:
@Caparow I've added default parameters & default object instance in next commit for convenience for most people who would be fine with container only working in tests - https://github.com/7mind/izumi/commit/df4e847132fcaf781b0089de088db3f4de8bb115

PR #1235, opened 2020-09-18
Fix #1015 move support modules for effect types from distage-framework to distage-core

### LINKED ISSUE #1015: Move effect modules from distage-framework to distage-core
Effect modules are useful most of the time when using distage, even if using injector directly without the framework, so should be available earlier, right in distage-core.

neko-kai:


distage/distage-core/src/main/scala/izumi/distage/effect/modules/MonixBIODIEffectModule.scala
```diff
@@ -0,0 +1,37 @@
+package izumi.distage.effect.modules
+
+import cats.Parallel
+import cats.effect.{ConcurrentEffect, ContextShift, Timer}
```
VladPodilnyk: Is it OK that this file is identical to `MonixDIEffectModule.scala`?

neko-kai: @VladPodilnyk this is for `monix.bio.IO` from `monix-bio`, while that one is for `monix.eval.Task` from `monix`. These are for different libraries.

VladPodilnyk: Oh, yeah, sorry, I've not spotted that imports are different.

PR #1251, opened 2020-09-25
Feature/memoization levels

### LINKED ISSUE #1188: distage-testkit: Generalize envs & memoizationRoots to multi-leveled memoization roots
Right now there's only one level of memoization in tests, either all of the components in `memoizationRoots` are shared between tests or _none_ are and a new env is created where ALL the components are recreated from scratch.

This poses a problem for components that are more extremely heavy to initialize and are more "global" than others, e.g. – docker containers. Currently the `reuse` mechanism in `distage-framework-docker` is abused to allow global memoization of docker containers, even though initially it was only supposed to allow memoization between independent consecutive test runs, not in the same test run between parallel envs.

However, what we can do is allow definition of Multi-leveled memoization roots and to support these we'll have to make the current memoization env mechanism recursive – partiiton into a set of envs by first level of memoization roots, then merge by locator results, then repeat the same procedure for the next sets of memoization roots. This would allow definitions such as:

```scala
def config = TestConfig(
  memoizationRoots = MemoizationLevels(
    0 -> Set(DIKey[PostgresDocker]),
    1 -> Set(DIKey[Transactor]),
    2 -> Set(DIKey[Repo1], DIKey[Repo2]),
  )
```

API not decided, levels may be represented with a different structure, the above is for illustration.

/cc @Caparow

Caparow:
Closes #1188


distage/distage-core-api/src/main/scala/izumi/distage/model/definition/DIResource.scala
```diff
@@ -554,15 +554,15 @@ object DIResource {
       evalMap[G, OuterResource](a => DIEffect[G].map(f(a))(_ => a))
 
     /** Wrap acquire action of this resource in another effect, e.g. for logging purposes */
-    final def wrapAcquire[G[x] >: F[x]](f: (=> G[InnerResource]) => G[InnerResource]): DIResourceBase[G, OuterResource] =
-      wrapAcquireImpl[G, OuterResource](this: this.type)(f)
+    final def wrapAcquire[G[x] >: F[x], I2 >: InnerResource](f: (=> G[InnerResource]) => G[I2]): DIResourceBase[G, OuterResource] =
```
neko-kai: This is unsound, e.g.  subsequent `release` will throw `ClassCastException` if `I2`=Any

Caparow: @neko-kai Yup, I've reverted these changes. Was trying to deal with DIResource in tree traverse with action wrapping, but it was overcomplicated.

Caparow:
Current memoized tree render:
image

//cc @neko-kai @pshirshov

neko-kai:
🎉

PR #1257, opened 2020-09-30
Add Tele2 to users

mehakun:


neko-kai:
@mehakun 
Thank you! 🙏

PR #1260, opened 2020-10-03
WIP: integrate sbt-doctest

neko-kai:
Current blocker:

* `sbt-doctest` generates test files inside the current artifact's test scope, but e.g. code examples in `distage-core-api` forward-reference classes that are only available in `distage-core` and onward — they cannot compile in `distage-core-api / test` scope. Now, this parameter the directory to dump source files to is controlled by this line https://github.com/tkawachi/sbt-doctest/blob/master/src/main/scala/com/github/tkawachi/doctest/DoctestPlugin.scala#L103, if we provide an alternative definition for `doctestGenTests` that uses a different key instead of `managedSourceDirectories in Test`, we could override that key and dump the tests into a different directory (e.g. `microsite` - where all the libraries are available at the same time)

/cc @vpodlnk @VladPodilnyk - would you be interested in exploring this task?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3065,7 +3065,9 @@ lazy val `izumi` = (project in file("."))
     scmInfo in ThisBuild := Some(ScmInfo(url("https://github.com/7mind/izumi"), "scm:git:https://github.com/7mind/izumi.git")),
     scalacOptions in ThisBuild ++= Seq(
       s"-Xmacro-settings:scalatest-version=${V.scalatest}"
-    )
+    ),
+    doctestOnlyCodeBlocksMode in ThisBuild := true,
+    doctestTestFramework in ThisBuild := DoctestTestFramework.ScalaTest
   )
   .disablePlugins(AssemblyPlugin)
   .aggregate(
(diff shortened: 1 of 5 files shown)


neko-kai:
`distage-core-api/test`

VladPodilnyk:
> Current blocker:
> 
>     * `sbt-doctest` generates test files inside the current artifact's test scope, but e.g. code examples in `distage-core-api` forward-reference classes that are only available in `distage-core` and onward — they cannot compile in `distage-core-api / test` scope. Now, this parameter the directory to dump source files to is controlled by this line https://github.com/tkawachi/sbt-doctest/blob/master/src/main/scala/com/github/tkawachi/doctest/DoctestPlugin.scala#L103, if we provide an alternative definition for `doctestGenTests` that uses a different key instead of `managedSourceDirectories in Test`, we could override that key and dump the tests into a different directory (e.g. `microsite` - where all the libraries are available at the same time)
> 
> 
> /cc @vpodlnk @VladPodilnyk - would you be interested in exploring this task?

Jeez, I'm so sorry, somehow, I've just missed a notification about this 🤦 .  I didn't ignore this, just missed it out, sorry.
Super late answer, but I thought it won't be polite if I leave it like it is.

neko-kai:
@VladPodilnyk no worries!

PR #1278, opened 2020-10-06
Feature: Free monad

Caparow:


fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/data/FreeApplicative.scala
```diff
@@ -0,0 +1,10 @@
+package izumi.functional.bio.data
+
+/** Scenario applicative is needed to create Pure scenario instance for [[Free.map]] implementation. */
```
neko-kai: scenario comment

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/data/FreeInterpreter.scala
```diff
@@ -0,0 +1,20 @@
+package izumi.functional.bio.data
+
+/**
+  * Should executed scenario steps in a single context and save all scenario metrics.
```
neko-kai: scenario comment

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/data/FreeInterpreter.scala
```diff
@@ -0,0 +1,20 @@
...
+  */
+trait FreeInterpreter[F[+_, +_], S[_[+_, +_], _, _], Scope] {
+  /** Single step execution. */
+  def interpret[E, A](scope: Scope, s: S[F, E, A]): F[Nothing, FreeInterpreter.Result[F, S, Scope]]
```
neko-kai: The scope parameter may be implemented on the level of interpreter for `F` and removed from the base definition?

neko-kai: We seem to be missing an equivalent to `foldMap`[[1]](https://github.com/typelevel/cats/blob/master/free/src/main/scala/cats/free/Free.scala#L158)[[2]](https://typelevel.org/cats/datatypes/freemonad.html#run-your-program)
That is a function like
```scala
def foldMap[F[+_, +_], G[+_, +_], E, A](f: Free[F, E, A])(g: F ~> G)(implicit G: BIOPanic[G]): G[E, A]
```
That would convert inner `F` in Free to G and convert the branches of Free to G, thereby converting everything to G

neko-kai:
let's rename this `FreeBracket` and make a smaller copy called `Free` with just the monad ops and merge 👍

Caparow:
@neko-kai I've added small `Free` with BIOMonad instance. 
Also, I've tried to rename old `Free` -> `FreePanic` and add `BIOPanic[FreePanic[S, +?, +?]]` instance, but it appears to be complicated. Fail and Termination branches for some reason break the trampoline. (`BIOPanic` instance is needed for `mapK` implementation)

neko-kai:
@Caparow hmm, i wouldn't know what's the issue here, you may push failing code here or in a different branch so i may look at it though

Caparow:
@neko-kai I was able to add `BIOPanic` instance. The trampoline was breaking due to trait usages instead of wrapping classes :disappointed:. 
Also, I needed to wrap the function into `Pure` to avoid adding applicative for `S[_, _]` type here:
```scala
    @inline override def fromEither[E, V](effect: => Either[E, V]): FreePanic[S, E, V] = FreePanic.pure(() => effect).flatMap {
      _() match {
        case Left(value) => fail(value)
        case Right(value) => pure(value)
      }
    }
```
What do you think?

neko-kai:
@Caparow 
`FreePanic.pure(() => effect).flatMap { _()` is for suspension, right? Maybe better to write `FreePanic.unit.flatMap { effect() match { ... } }` because you don't actually need the lambda value `() => effect`

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/data/FreePanic.scala
```diff
@@ -0,0 +1,146 @@
...
+
+  @inline final def redeem[B, E1](err: E => FreePanic[S, E1, B], succ: A => FreePanic[S, E1, B]): FreePanic[S, E1, B] =
+    FreePanic.Redeem(this, err, succ)
+  @inline final def redeemPure[B](err: E => B, succ: A => B): FreePanic[S, Nothing, B] =
+    FreePanic.Redeem[S, E, Nothing, A, B](this, e => FreePanic.pure(err(e)), a => FreePanic.pure(succ(a)))
+  @inline final def catchAll[A1 >: A, E1](err: E => FreePanic[S, E1, A1]): FreePanic[S, E1, A1] =
+    FreePanic.Redeem(this, err, (a: A) => FreePanic.pure(a))
+  @inline final def catchSome[A1 >: A, E1 >: E](err: PartialFunction[E, FreePanic[S, E1, A1]]): FreePanic[S, E1, A1] =
+    FreePanic.Redeem(this, err.orElse[E, FreePanic[S, E1, A1]] { case e => FreePanic.fail(e) }, (a: A) => FreePanic.pure(a))
+  @inline final def swap: FreePanic[S, A, E] =
```
neko-kai: Let's keep the same names everywhere.
```suggestion
  @inline final def flip: FreePanic[S, A, E] =
```

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/data/FreePanic.scala
```diff
@@ -0,0 +1,146 @@
...
+import scala.util.Try
+
+sealed trait FreePanic[S[_, _], +E, +A] {
+  @inline final def flatMap[B, E1 >: E](fun: A => FreePanic[S, E1, B]): FreePanic[S, E1, B] = FreePanic.FlatMapped[S, E, E1, A, B](this, fun)
+  @inline final def map[B](fun: A => B): FreePanic[S, E, B] = flatMap(a => FreePanic.pure[S, B](fun(a)))
+  @inline final def as[B](as: B): FreePanic[S, E, B] = map(_ => as)
+  @inline final def *>[B, E1 >: E](sc: FreePanic[S, E1, B]): FreePanic[S, E1, B] = flatMap(_ => sc)
+  @inline final def <*[B, E1 >: E](sc: FreePanic[S, E1, B]): FreePanic[S, E1, A] = flatMap(sc.as)
+
+  @inline final def sandbox: FreePanic[S, BIOExit.Failure[E], A] =
```
neko-kai, line 14: Let's keep the same names everywhere.
```suggestion
  @inline final def sandboxExit: FreePanic[S, BIOExit.Failure[E], A] =
```

Caparow:
> @Caparow
> `FreePanic.pure(() => effect).flatMap { _()` is for suspension, right? Maybe better to write `FreePanic.unit.flatMap { effect() match { ... } }` because you don't actually need the lambda value `() => effect`

Yup, you are right. Will update.

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/data/FreePanic.scala
```diff
@@ -0,0 +1,146 @@
+package izumi.functional.bio.data
+
+import izumi.functional.bio.{BIOExit, BIOPanic}
+
+import scala.util.Try
+
+sealed trait FreePanic[S[_, _], +E, +A] {
```
neko-kai: Hmm, I feel a bit sad that the `S` is invariant, but if we change it to `FreePanic[+S[_, _], +E, +A]` this will cause every method to have a noisy `[S1[e, a] >: S[e, a]` parameter.
Could be worth it tho, it would allow us to auto-widen to a union of functors when we get unions (in a couple of years...) or if we use a tagless union encoding (https://medium.com/@odomontois/tagless-unions-in-scala-2-12-55ab0100c2ff)

Caparow: @neko-kai I've updated S type to be covariant for both `Free` and `FreePanic`. Don't sure if it's really needing but why not.

neko-kai:
@Caparow 🎉 !

PR #1285, opened 2020-10-08
Fix #1226 New compile-time checker API

### LINKED ISSUE #1226: distage: New API for compile-time checker
After https://github.com/7mind/izumi/pull/1194 compile-time checking is now seriously usable and must be given an overhaul to the API. Current proof-of-concept API is extremely hard to find & use, we need to replace it with an API that:

1. automatically share settings with the RoleLauncher - you should be able to not copy configuration between compile-time checker & role launcher at least in simple cases
2. automatically create distage-testkit tests – these tests should do a compile-time check when compiled & the same check at runtime when running[? is the additional runtime check useful, maybe just a switch compile-time-only/runtime-only ? - additional runtime check is useful to be able to more easily attach a debugger and inspect values. It's harder to attach debugger to the compiler].

neko-kai:


neko-kai:
Merged in WIP state to allow a milestone release of other features. Continued in https://github.com/7mind/izumi/pull/1296

Issue #1309, opened 2020-11-09
distage.Lifecycle: add `catchAll` and other error handling methods


neko-kai:

neko-kai:
Done in #1422

PR #1422, opened 2021-03-25
New methods for Lifecycle: catchAll + foldM

VladPodilnyk:
Seems OK.
Feel free to throw tomatoes at me xD
Partially addresses https://github.com/7mind/izumi/issues/1309

### DIFF
--- a/distage/distage-core-api/src/main/scala/izumi/distage/model/definition/Lifecycle.scala
+++ b/distage/distage-core-api/src/main/scala/izumi/distage/model/definition/Lifecycle.scala
@@ -999,9 +999,9 @@ object Lifecycle extends LifecycleCatsInstances {
       override def acquire: F[InnerResource] = F.maybeSuspend(new AtomicReference(Nil))
       override def release(finalizers: InnerResource): F[Unit] = F.suspendF(F.traverse_(finalizers.get())(_.apply()))
       override def extract[C >: B](finalizers: InnerResource): Either[F[C], C] = Left {
-        bracketAppendFinalizer(finalizers)(self) {
-          inner1: self.InnerResource =>
-            F.definitelyRecover {
+        F.definitelyRecover {
+          bracketAppendFinalizer(finalizers)(self) {
+            inner1: self.InnerResource =>
               self.extract(inner1).fold(_.map(success), F pure success(_)).flatMap {
                 that: Lifecycle[F, B] =>
                   bracketAppendFinalizer(finalizers)(that) {
@@ -1009,13 +1009,13 @@ object Lifecycle extends LifecycleCatsInstances {
                       that.extract[C](inner2).fold(identity, F.pure)
                   }
               }
-            } {
-              error =>
-                val onError: Lifecycle[F, B] = failure(error)
-                bracketAppendFinalizer(finalizers)(onError) {
-                  inner3: onError.InnerResource =>
-                    onError.extract[C](inner3).fold(identity, F.pure)
-                }
+          }
+        } {
+          error =>
+            val onError: Lifecycle[F, B] = failure(error)
+            bracketAppendFinalizer(finalizers)(onError) {
+              inner3: onError.InnerResource =>
+                onError.extract[C](inner3).fold(identity, F.pure)
             }
         }
       }
--- a/distage/distage-core-api/src/main/scala/izumi/distage/model/definition/Lifecycle.scala
+++ b/distage/distage-core-api/src/main/scala/izumi/distage/model/definition/Lifecycle.scala
@@ -258,6 +258,8 @@ trait Lifecycle[+F[_], +OuterResource] {
 
   final def void[G[x] >: F[x]: QuasiFunctor]: Lifecycle[G, Unit] = map[G, Unit](_ => ())
 
+  final def catchAll[G[x] >: F[x], E <: Throwable, B](f: E => Lifecycle[G, B]): Lifecycle[G, B] = ???
+
   @inline final def widen[B >: OuterResource]: Lifecycle[F, B] = this
   @inline final def widenF[G[x] >: F[x]]: Lifecycle[G, OuterResource] = this
 }
@@ -961,6 +963,9 @@ object Lifecycle extends LifecycleCatsInstances {
       override def extract[B >: A](resource: InnerResource): Either[F[B], B] = self.extract(resource)
     }
   }
+
+  @inline
+  private final def foldMImpl[F[_], E, A, B](failure: E => Lifecycle[F, B], success: A => Lifecycle[F, B]): Lifecycle[F, B] = {}
 
   @deprecated("renamed to AdaptFunctoid", "1.0")
   type AdaptProvider[A] = AdaptFunctoidImpl[A]
--- a/distage/distage-core-api/src/main/scala/izumi/distage/model/definition/Lifecycle.scala
+++ b/distage/distage-core-api/src/main/scala/izumi/distage/model/definition/Lifecycle.scala
@@ -968,7 +968,7 @@ object Lifecycle extends LifecycleCatsInstances {
   }
 
   @inline
-  private final def foldMImpl[F[_], A, B >: A](
+  private final def foldMImpl[F[_], A, B](
     self: Lifecycle[F, A]
   )(failure: Throwable => Lifecycle[F, B],
     success: A => Lifecycle[F, B],
--- a/distage/distage-core-api/src/main/scala/izumi/distage/model/definition/Lifecycle.scala
+++ b/distage/distage-core-api/src/main/scala/izumi/distage/model/definition/Lifecycle.scala
@@ -999,25 +999,28 @@ object Lifecycle extends LifecycleCatsInstances {
       override def acquire: F[InnerResource] = F.maybeSuspend(new AtomicReference(Nil))
       override def release(finalizers: InnerResource): F[Unit] = F.suspendF(F.traverse_(finalizers.get())(_.apply()))
       override def extract[C >: B](finalizers: InnerResource): Either[F[C], C] = Left {
-        F.definitelyRecover {
-          bracketAppendFinalizer(finalizers)(self) {
-            inner1: self.InnerResource =>
-              self.extract(inner1).fold(_.map(success), F pure success(_)).flatMap {
-                that: Lifecycle[F, B] =>
-                  bracketAppendFinalizer(finalizers)(that) {
-                    inner2: that.InnerResource =>
-                      that.extract[C](inner2).fold(identity, F.pure)
-                  }
-              }
-          }
-        } {
-          error =>
-            val onError: Lifecycle[F, B] = failure(error)
-            bracketAppendFinalizer(finalizers)(onError) {
-              inner3: onError.InnerResource =>
-                onError.extract[C](inner3).fold(identity, F.pure)
-            }
+        val action = bracketAppendFinalizer(finalizers)(self) {
+          inner: self.InnerResource =>
+            self.extract(inner).fold(identity, F.pure)
         }
+
+        val onSuccess = (a: A) => {
+          val resource = success(a)
+          bracketAppendFinalizer(finalizers)(resource) {
+            inner: resource.InnerResource =>
+              resource.extract[C](inner).fold(identity, F.pure)
+          }
+        }
+
+        val onError = (err: Throwable) => {
+          val resource = failure(err)
+          bracketAppendFinalizer(finalizers)(resource) {
+            inner: resource.InnerResource =>
+              resource.extract[C](inner).fold(identity, F.pure)
+          }
+        }
+
+        F.redeem(action)(onError, onSuccess)
       }
     }
   }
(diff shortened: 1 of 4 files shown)


distage/distage-core-api/src/main/scala/izumi/distage/model/definition/Lifecycle.scala
```diff
@@ -240,6 +240,8 @@ trait Lifecycle[+F[_], +OuterResource] {
   final def evalMap[G[x] >: F[x]: QuasiIO, B](f: OuterResource => G[B]): Lifecycle[G, B] = evalMapImpl[G, OuterResource, B](this)(f)
   final def evalTap[G[x] >: F[x]: QuasiIO](f: OuterResource => G[Unit]): Lifecycle[G, OuterResource] =
     evalMap[G, OuterResource](a => QuasiIO[G].map(f(a))(_ => a))
+  final def foldM[G[x] >: F[x]: QuasiIO, B](onFailure: Throwable => Lifecycle[G, B], onSuccess: OuterResource => Lifecycle[G, B]): Lifecycle[G, B] =
```
neko-kai: Rename just to be consistent with `izumi.functional.bio.Error3#redeem`:
```suggestion
  final def redeem[G[x] >: F[x]: QuasiIO, B](onFailure: Throwable => Lifecycle[G, B], onSuccess: OuterResource => Lifecycle[G, B]): Lifecycle[G, B] =
```

distage/distage-core-api/src/main/scala/izumi/distage/model/definition/Lifecycle.scala
```diff
@@ -962,6 +967,61 @@ object Lifecycle extends LifecycleCatsInstances {
...
+          use = use
+        )
+      }
+
+      override def acquire: F[InnerResource] = F.maybeSuspend(new AtomicReference(Nil))
+      override def release(finalizers: InnerResource): F[Unit] = F.suspendF(F.traverse_(finalizers.get())(_.apply()))
+      override def extract[C >: B](finalizers: InnerResource): Either[F[C], C] = Left {
+        bracketAppendFinalizer(finalizers)(self) {
+          inner1: self.InnerResource =>
+            F.definitelyRecover {
```
neko-kai: This will not run the failure handler if `self.acquire` failed. Current `Lifecycle` encoding is roughly similar to [old pre-RC19 ZManaged encoding](https://github.com/zio/zio/blob/v1.0.0-RC18/core/shared/src/main/scala/zio/ZManaged.scala) (where `Lifecycle#extract` = `Reservation#acquire`, `Lifecycle#acquire` = `ZManaged#reservation`), you may be able to look at their implementation of [foldM](https://github.com/zio/zio/blob/v1.0.0-RC18/core/shared/src/main/scala/zio/ZManaged.scala#L396) for tips.

VladPodilnyk: Yeah, I've taken inspiration there, but seems I misunderstood some parts. Thanks for the review and you notes about implementation!

VladPodilnyk: Done.

VladPodilnyk:
I guess, next step is to make Lifecycle with F[_, _] which gives ability to work with different error types.
This seems like major change or not ?

distage/distage-core-api/src/main/scala/izumi/distage/model/definition/Lifecycle.scala
```diff
@@ -962,6 +967,61 @@ object Lifecycle extends LifecycleCatsInstances {
...
+
+      override def acquire: F[InnerResource] = F.maybeSuspend(new AtomicReference(Nil))
+      override def release(finalizers: InnerResource): F[Unit] = F.suspendF(F.traverse_(finalizers.get())(_.apply()))
+      override def extract[C >: B](finalizers: InnerResource): Either[F[C], C] = Left {
+        F.definitelyRecover {
+          bracketAppendFinalizer(finalizers)(self) {
+            inner1: self.InnerResource =>
+              self.extract(inner1).fold(_.map(success), F pure success(_)).flatMap {
+                that: Lifecycle[F, B] =>
+                  bracketAppendFinalizer(finalizers)(that) {
```
neko-kai: @VladPodilnyk It is possible for `failure` and `success` to both run at the same time with this scheme, if the failure occurs at or after this line, which is logically incorrect.

VladPodilnyk: @neko-kai, now should be fixed. I have some issues with Lifecycle instance for Identity. It does not work now, since action is eagerly evaluated.

distage/distage-core-api/src/main/scala/izumi/distage/model/effect/QuasiIO.scala
```diff
@@ -51,6 +51,8 @@ trait QuasiIO[F[_]] extends QuasiApplicative[F] {
     */
   def definitelyRecoverCause[A](action: => F[A])(recoverCause: (Throwable, (() => Throwable)) => F[A]): F[A]
 
+  def redeem[A, B](action: F[A])(failure: Throwable => F[B], success: A => F[B]): F[B]
```
neko-kai: @VladPodilnyk Are you sure a separate `redeem` method is required when it's equivalent to `action.definitelyRecover(failure).flatMap(success)`?

distage/distage-core-api/src/main/scala/izumi/distage/model/definition/Lifecycle.scala
```diff
@@ -962,6 +967,64 @@ object Lifecycle extends LifecycleCatsInstances {
...
+      override def acquire: F[InnerResource] = F.maybeSuspend(new AtomicReference(Nil))
+      override def release(finalizers: InnerResource): F[Unit] = F.suspendF(F.traverse_(finalizers.get())(_.apply()))
+      override def extract[C >: B](finalizers: InnerResource): Either[F[C], C] = Left {
+        val action = bracketAppendFinalizer(finalizers)(self) {
+          inner: self.InnerResource =>
+            self.extract(inner).fold(identity, F.pure)
+        }
+
+        val onSuccess = (a: A) => {
+          val resource = success(a)
```
neko-kai, line 1008: This extraction into a val cannot be done safely due to `Identity` instance

distage/distage-core-api/src/main/scala/izumi/distage/model/definition/Lifecycle.scala
```diff
@@ -962,6 +967,64 @@ object Lifecycle extends LifecycleCatsInstances {
...
+    self: Lifecycle[F, A]
+  )(failure: Throwable => Lifecycle[F, B],
+    success: A => Lifecycle[F, B],
+  )(implicit F: QuasiIO[F]
+  ): Lifecycle[F, B] = {
+    import QuasiIO.syntax._
+    new Lifecycle[F, B] {
+      override type InnerResource = AtomicReference[List[() => F[Unit]]]
+
+      private[this] def bracketAppendFinalizer[a, b](finalizers: InnerResource)(lifecycle: Lifecycle[F, a])(use: lifecycle.InnerResource => F[b]): F[b] = {
```
neko-kai, line 981:
```suggestion
      private[this] def bracketAppendFinalizer[a, b](finalizers: InnerResource)(lifecycle: => Lifecycle[F, a])(use: lifecycle.InnerResource => F[b]): F[b] = {
```

Must suspend here to ensure that any effects in construction of lifecycle are wrapped into `bracket`'s `try` (in Identity instance). (Lifecycle/QuasiIO must work correctly with impure code)

neko-kai: Hmm, this may be tricky because of path dependency

distage/distage-core-api/src/main/scala/izumi/distage/model/effect/QuasiIO.scala
neko-kai:
```suggestion
  def redeem[A, B](action: => F[A])(failure: Throwable => F[B], success: A => F[B]): F[B]
```
Also you would need suspension here to make it work with `Identity`.

distage/distage-core-api/src/main/scala/izumi/distage/model/definition/Lifecycle.scala
```diff
@@ -962,6 +967,64 @@ object Lifecycle extends LifecycleCatsInstances {
...
+          }
+        )(release = _ => F.unit)(
+          use = use
+        )
+      }
+
+      override def acquire: F[InnerResource] = F.maybeSuspend(new AtomicReference(Nil))
+      override def release(finalizers: InnerResource): F[Unit] = F.suspendF(F.traverse_(finalizers.get())(_.apply()))
+      override def extract[C >: B](finalizers: InnerResource): Either[F[C], C] = Left {
+        val action = bracketAppendFinalizer(finalizers)(self) {
```
neko-kai: `val action` must be inlined or become `def action`, because with `F=Identity` it's not an `action` but a `result`

distage/distage-core-api/src/main/scala/izumi/distage/model/definition/Lifecycle.scala
neko-kai: @VladPodilnyk 
This seems to work as an encoding that supports suspension and path dependency:
```scala
private[this] sealed trait LifecyclePair[+a] {
  val lifecycle: Lifecycle[F, a]
  val inner: lifecycle.InnerResource
}
private[this] def LifecyclePair[a](lifecycle0: Lifecycle[F, a])(inner0: lifecycle0.InnerResource): LifecyclePair[a] = new LifecyclePair[a] {
  override val lifecycle: lifecycle0.type = lifecycle0
  override val inner: lifecycle.InnerResource = inner0
}

private[this] def bracketAppendFinalizer[a, b, t](
  finalizers: InnerResource
)(lifecycleCtor: () => Lifecycle[F, a]
)(use: LifecyclePair[a] => F[b]
): F[b] = {
  F.bracket(
    acquire = {
      val lifecycle = lifecycleCtor()
      lifecycle.acquire.flatMap {
        a =>
          F.maybeSuspend {
            // can't use `.updateAndGet` because of Scala.js
            var oldValue = finalizers.get()
            while (!finalizers.compareAndSet(oldValue, (() => lifecycle.release(a)) :: oldValue)) {
              oldValue = finalizers.get()
            }
            LifecyclePair(lifecycle)(a)
          }
      }
    }
  )(release = _ => F.unit)(
    use = use
  )
}
```

distage/distage-core-api/src/main/scala/izumi/distage/model/definition/Lifecycle.scala
neko-kai: @VladPodilnyk I think I don't see any issues now, seems like the only thing left is a few by-names to accomodate `Identity`

distage/distage-core-api/src/main/scala/izumi/distage/model/effect/QuasiIO.scala
VladPodilnyk: >@VladPodilnyk Are you sure a separate redeem method is required when it's equivalent to action.definitelyRecover(failure).flatMap(success)?

The problem is action has type `F[A]`, failure is a function from `Throwable to F[B]` hence I couldn't use `definatelyRecover` since it requires the same type for action and failure. So I decided to add some functionality.

neko-kai: Right, I wasn't correct here

distage/distage-core-api/src/main/scala/izumi/distage/model/definition/Lifecycle.scala
VladPodilnyk: Do you propose to extract it into a pattern ☝️ ?
(it's not clear for me what problem does it solves)
Also, we probably should come up with another name bc this one is not very informative, IMO.

neko-kai: @VladPodilnyk 
> (it's not clear for me what problem does it solves)

You need the `lifecycle` parameter to be by-name because it may produce effects immediately when called, instead of when executed (it's natural to perform effects inplace if you use impure Identity instance)

But if you write it as by-name, it's now impossible to write the `lifecycle.InnerResource` type:
```scala
private[this] def bracketAppendFinalizer[a, b](finalizers: InnerResource)(lifecycle: => Lifecycle[F, a])(use: lifecycle.InnerResource => F[b]): F[b] = {
```

This type will produce an error because `lifecycle` is now by-name - every time you refer to it in code, the by-name parameter will run again and produce a new value (operationally `=> Lifecycle[F, a]` is exactly the same as `() => Lifecycle[F, a]`). 
Because it produces a new lifecycle each time, you cannot refer to `lifecycle.InnerResource` — a type member of `lifecycle` value — because there is no "`lifecycle` value" here, only a function that makes new Lifecycles. So we package the result "value" into a trait to make it referable (the reason it's not a class is because classes do not support path-dependency in Scala 2, only traits)

In short, try to implement the above signature and you'll be able to experience the limitations that force it to be this way.

neko-kai: Right now you need to get rid of `val` extraction and write this thing as if everything is imperative (because it IS when using `Identity`) and then it's good to merge 🚀

VladPodilnyk: @neko-kai, thanks for this excellent explanation!

distage/distage-core-api/src/main/scala/izumi/distage/model/definition/Lifecycle.scala
VladPodilnyk: Replaced with `def` now.

distage/distage-core-api/src/main/scala/izumi/distage/model/definition/Lifecycle.scala
VladPodilnyk: Done.

neko-kai (approved):

neko-kai:
@VladPodilnyk Thank you! 🙏

VladPodilnyk:
:hugs:

PR #1445, opened 2021-04-16
BIO instances for Either

VladPodilnyk:
TBH, I don't know if understand it correctly (pls, don't send me to exile if I did a stupid thing 🙈). Anyway, I couldn't imagine situation when somebody will require such instances.

UPD: Tests coming....

### DIFF
--- a/fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/impl/BioEither.scala
+++ b/fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/impl/BioEither.scala
@@ -7,25 +7,26 @@ import scala.util.Try
 object BioEither extends BioEither
 
 class BioEither extends Error2[Either] {
-  override def pure[A](a: A): Either[Nothing, A] = Right(a)
-  override def map[R, E, A, B](r: Either[E, A])(f: A => B): Either[E, B] = r.map(f)
+  @inline override def pure[A](a: A): Either[Nothing, A] = Right(a)
+  @inline override def map[R, E, A, B](r: Either[E, A])(f: A => B): Either[E, B] = r.map(f)
 
   /** execute two operations in order, map their results */
-  override def map2[R, E, A, B, C](firstOp: Either[E, A], secondOp: => Either[E, B])(f: (A, B) => C): Either[E, C] = {
+  @inline override def map2[R, E, A, B, C](firstOp: Either[E, A], secondOp: => Either[E, B])(f: (A, B) => C): Either[E, C] = {
     firstOp.flatMap(a => secondOp.map(b => f(a, b)))
   }
-  override def flatMap[R, E, A, B](r: Either[E, A])(f: A => Either[E, B]): Either[E, B] = r.flatMap(f)
+  @inline override def flatMap[R, E, A, B](r: Either[E, A])(f: A => Either[E, B]): Either[E, B] = r.flatMap(f)
 
-  override def catchAll[R, E, A, E2](r: Either[E, A])(f: E => Either[E2, A]): Either[E2, A] = r.left.flatMap(f)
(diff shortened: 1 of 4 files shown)


neko-kai:
@VladPodilnyk 
> Also we couldn't go further than Monad2.

It should be possible to support up to `Error2` for Either – there is only one weirdness in that a `Guarantee2[Either]` instance is quite useless, it could be made a bit more useful in practice if the second parameter of `guarantee` is made by-name, although in pure context it hardly matters – `guarantee` & `bracket` on Either are useless because the only effect of Either is the ability to return an error and these two functions prohibit errors on `cleanup`/`release` making them no-ops.

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/impl/BioEither.scala
```diff
@@ -0,0 +1,31 @@
...
+
+  /** execute two operations in order, map their results */
+  override def map2[R, E, A, B, C](firstOp: Either[E, A], secondOp: => Either[E, B])(f: (A, B) => C): Either[E, C] = {
+    firstOp.flatMap(a => secondOp.map(b => f(a, b)))
+  }
+  override def flatMap[R, E, A, B](r: Either[E, A])(f: A => Either[E, B]): Either[E, B] = r.flatMap(f)
+
+  override def catchAll[R, E, A, E2](r: Either[E, A])(f: E => Either[E2, A]): Either[E2, A] = r.left.flatMap(f)
+  override def catchSome[R, E, A, E1 >: E](r: Either[E, A])(f: PartialFunction[E, Either[E1, A]]): Either[E1, A] = r.left.flatMap(f)
+  override def fail[E](v: => E): Either[E, Nothing] = Left(v)
```
VladPodilnyk: catchSome - seems weird, but the rest should be OK

VladPodilnyk:
> @VladPodilnyk
> 
> > Also we couldn't go further than Monad2.
> 
> It should be possible to support up to `Error2` for Either – there is only one weirdness in that a `Guarantee2[Either]` instance is quite useless, it could be made a bit more useful in practice if the second parameter of `guarantee` is made by-name, although in pure context it hardly matters – `guarantee` & `bracket` on Either are useless because the only effect of Either is the ability to return an error and these two functions prohibit errors on `cleanup`/`release` making them no-ops.

Aha, understood. Gurantee3 really stops me at first.

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/impl/BioEither.scala
neko-kai: @VladPodilnyk `catchSome` is incorrect, you should use `.applyOrElse` instead of passing f itself (it will throw exception when used as a function, instead it should do nothing if the error doesn't match the partial function)

neko-kai: Note: cats law tests will probably show the error in `catchSome` (we use it to implement `cats.ApplicativeError#recoverWith`)

VladPodilnyk:
Tests fails bc now we always have an instance of BIO for Either. So if we don't have Monix or ZIO in a classpath, we will find BioEither instance.
[click here](https://github.com/7mind/izumi/blob/d96cc177654ccd5cf7b750a24295ecb1b3ef4723/distage/distage-extension-config/src/test/scala_2.13/izumi/distage/impl/OptionalDependencyTest213.scala#L15)

Can I somehow require anything but BioEither? On the other hand I could simply do the following :
```scala
assert(new optSearch2[Functor2].find == BioEither)
```
Dunno, what is more preferable for you guys.
cc: @neko-kai

neko-kai:
@VladPodilnyk 
> On the other hand I could simply do the following :

That will work 👍

neko-kai (approved):

PR #1464, opened 2021-04-26
Feature: healthcheck for docker container exit

Caparow:


distage/distage-framework-docker/src/main/scala/izumi/distage/docker/healthcheck/ContainerExitedCheck.scala
```diff
@@ -0,0 +1,25 @@
...
+
+final class ContainerExitedCheck[Tag](canBeDestroyed: Boolean) extends ContainerHealthCheck[Tag] {
+  override def check(logger: IzLogger, container: DockerContainer[Tag], state: Docker.ContainerState): HealthCheckResult = {
+    state match {
+      case ContainerState.Running =>
+        logger.debug(s"$container still running, marked as unavailable.")
+        HealthCheckResult.Unavailable
+      case ContainerState.SuccessfullyExited =>
+        logger.debug(s"$container successfully exited, marked available.")
+        HealthCheckResult.Available
```
neko-kai: @Caparow @pshirshov 
What do you think about renaming `HealthCheckResult.Available/Unavailable` to `.Good`/`.Bad`? This check reverses meaning of `Available`/`Unavailable` which makes it harder to understand what's going on with the current names...

Caparow: @neko-kai Yeah, that's a good idea. I've updated namings, and a bit reworked pulling logic.
I'm curious if we need `alwaysPull` option now (or we can on `alwaysPull == true` pull images always, and if it's set `false` then test if image exists and pull if needed)? 

//cc @vilunov

neko-kai: @Caparow I think if we can test that image exists than we should just switch to that. `alwaysPull=false` was for working around the fact that we always did a `docker pull` just to make sure the image is available – but this causes unnecessary connections and is not required if we can check that the image is already downloaded.

Caparow: @neko-kai I've removed `alwaysPull` and added pulling timeout handling. (Also, I've added `F.fromEither` methods to not to throw so much errors.)

neko-kai (approved):

neko-kai:
@Caparow Thank you! 🙏

PR #1469, opened 2021-04-30
Feature: ipv6 in docker

Caparow:


fundamentals/fundamentals-platform/src/main/scala/izumi/fundamentals/platform/integration/PortCheck.scala
```diff
@@ -47,10 +50,18 @@ class PortCheck(timeout: FiniteDuration) {
...
 
   def checkPort(host: String, port: Int, clue: Option[String] = None): ResourceCheck = {
+    checkAddress(new InetSocketAddress(host, port), clue)
+  }
+
+  def checkAddressPort(address: InetAddress, port: Int, clue: Option[String] = None): ResourceCheck = {
+    checkAddress(new InetSocketAddress(address, port), clue)
+  }
+
+  def checkAddress(address: => InetSocketAddress, clue: Option[String] = None): ResourceCheck = {
```
neko-kai, line 54: why by-name address? It will always be evaluated anyway unless `new Socket` fails...

neko-kai (approved):
/cc @pshirshov

fundamentals/fundamentals-platform/src/main/scala/izumi/fundamentals/platform/integration/PortCheck.scala
Caparow: Yup, you are right. Removed it.

PR #1500, opened 2021-05-17
Add back hostV4 as a deprecated alias (silently broke compatibility in 1.0.6)

neko-kai:


distage/distage-framework-docker/src/main/scala/izumi/distage/docker/Docker.scala
```diff
@@ -25,7 +30,7 @@ object Docker {
     def address: InetAddress
     def host: String
 
-    override final def toString: String = address.getHostAddress
+    override final def toString: String = host
```
neko-kai, line 33: @Caparow I think it makes sense to print IPV6 addresses with the surrounding `[]` brackets, however I'm not sure if there was a reason to not do it this way in the first place ☝️

Caparow: @neko-kai I've thought of printing V6 addresses without any additional changes, just if it would be returned by InetAddress. But it's safe to print it with `[]`

Caparow (approved):

PR #1506, opened 2021-05-24
Return to old Fiber arity - `Fiber2` as the default, not `Fiber3` (mirrors Ref/Promise, lets Fiber3/2 for the same F to have compatible types)

neko-kai:


VladPodilnyk (approved):
Nice!


2.12 tests failed. Maybe we could remove covariance in `F` and write (I am not sure that this will help, just guessing 🙈 ) 
```scala
def x2[FR[-_, +_, +_], F[E, A] >: FR[Any, E, A] <: FR[Any, E, A]: Applicative2: Fork2]
```
error
```
covariant type E occurs in contravariant position in type [+E, +A] >: FR[Any,E,A] <: FR[Any,E,A] of type F
```

neko-kai:
@VladPodilnyk 
fixed, i think
> Maybe we could remove covariance in F

Unfortunately we require covariance everywhere so we wouldn't be able to write `Applicative2[F]` if we remove variance... but 2.12 is bad at variance so we can't write it the original way either, so we have to change the test itself

VladPodilnyk:
@neko-kai, understood 😉 
thx!

PR #1528, opened 2021-06-11
Bug: incorrect failure on missing tag

Caparow:


pshirshov:
@neko-kai : seems like we have two issues. `make` emits some code which incorrectly tries to combine tag for a proper type with an empty arg list. And reflect fails on that.

```scala
class MissingTagOfTheNestedTypeTest extends AnyWordSpec with MkInjector {
  "fail with no Tag found" in {
    moduleWithoutTag[NestedTpe.Int.type]
  }
}
object MissingTagOfTheNestedTypeTest {
  sealed trait NestedTpe {
    type TPE
  }
  object NestedTpe {
    case object Int extends NestedTpe {
      override type TPE = Int
    }
  }

  final case class Dep[N <: NestedTpe](tpe: N#TPE)

  def moduleWithoutTag[N <: NestedTpe: Tag]: ModuleDef = new ModuleDef {
    make[Dep[N]]
  }
}
```

neko-kai:
@Caparow This now fails to compile at all with latest izumi / izumi-reflect.
```scala
/home/runner/work/izumi/izumi/distage/distage-core/src/test/scala/izumi/distage/injector/MissingTagOfTheNestedTypeTest.scala:40:9:
could not find implicit value for izumi.reflect.Tag[N#TPE].
  Did you forget to put on a Tag, TagK or TagKK context bound on one of the parameters in N#TPE?
  e.g. def x[T: Tag, F[_]: TagK] = ...

Error:  
Error:  
Error:  <trace>: 
Error:    deriving Tag for N#TPE, dealiased: N#TPE:
Error:    could not find implicit value for Tag[N#TPE]: N#TPE is a type parameter without an implicit Tag!
Error:      make[Dep[N]]
```
Should this be closed?

PR #1550, opened 2021-07-28
Add Scheduler +  newtype trait

VladPodilnyk:
- "in house" scheduler
- a newtype trait  that provides capabilities of `estatico/newtype` lib (an opaque wrapper around a type)

### DIFF
--- a/fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/retry/RetryPolicy.scala
+++ b/fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/retry/RetryPolicy.scala
@@ -1,13 +1,13 @@
 package izumi.functional.bio.retry
 
 import izumi.functional.bio.retry.RetryPolicy.{ControllerDecision, RetryFunction}
-import izumi.functional.bio.{Applicative2, F}
+import izumi.functional.bio.{F, Monad2}
 
 import java.time.ZonedDateTime
 import java.util.concurrent.TimeUnit
 import scala.concurrent.duration.FiniteDuration
 
-abstract class RetryPolicy[F[+_, +_]: Applicative2, -A, +B](val action: RetryFunction[F, A, B]) {
+abstract class RetryPolicy[F[+_, +_]: Monad2, -A, +B](val action: RetryFunction[F, A, B]) {
   def &&[A1 <: A, B1](policy: RetryPolicy[F, A1, B1]): RetryPolicy[F, A1, (B, B1)] = {
     def combined(left: RetryFunction[F, A, B], right: RetryFunction[F, A1, B1]): RetryFunction[F, A1, (B, B1)] =
       (now: ZonedDateTime, in: A1) =>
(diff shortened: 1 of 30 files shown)


neko-kai:
@VladPodilnyk 2.12 build is failing with 
```
[error] /home/vsts/work/1/s/fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/impl/SchedulerMonix.scala:10:18: object DurationConverters is not a member of package jdk
[error] import scala.jdk.DurationConverters._
```
I think you need to add collection-compat dependency to fundamentals-bio to fix that

VladPodilnyk:
@neko-kai thanks for this note, I'll keep that in mind 🙂 
I was working only with  2.13, wanna make a full solution and then check 2.12 build.

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/retry/package.scala
```diff
@@ -0,0 +1,14 @@
...
+import java.time.{Instant, ZoneOffset, ZonedDateTime}
+
+package object retry {
+  type Scheduler3[F[-_, +_, +_]] = Scheduler2[F[Any, +_, +_]]
+  object Scheduler3 {
+    @inline def apply[F[-_, +_, +_]: Scheduler3]: Scheduler3[F] = implicitly
+  }
+
+  @inline def toZonedDateTime(epochMillis: Long): ZonedDateTime = {
+    ZonedDateTime.ofInstant(Instant.ofEpochMilli(epochMillis), ZoneOffset.UTC)
```
VladPodilnyk, line 13: I've moved it from `bio` package because it's kinda weird to have `Scheduler2` under `retry` and `Scheduler3` under bio package. Obviously, it would be awesome to store it with other type aliases, but
I think it makes sense to have a separate package for this one.

fundamentals/fundamentals-bio/src/test/scala/izumi/fundamentals/bio/test/SchedulerTest.scala
```diff
@@ -0,0 +1,157 @@
...
+import java.util.concurrent.TimeUnit
+import scala.concurrent.duration._
+
+class SchedulerTest extends AnyWordSpec {
+  //private val monixRunner = UnsafeRun2.createMonixBIO(TestScheduler(), bio.IO.defaultOptions)
+  private val zioRunner = UnsafeRun2.createZIO(Platform.default)
+
+  // TODO: monix requires timer instance which apparently I should make by myself since I cannot instantiate a default one?
+  //private val monixScheduler = new SchedulerMonix(cats.effect.Timer)
+  private val zioTestClock = zio.Has(zio.clock.Clock.Service.live)
```
VladPodilnyk: @neko-kai, I haven't found other way to test Scheduler except to use runner to run all effects.
I have not problem with ZIO instances, but I am kinda struggling with Monix, since I cannot access a Timer that is used in IOApp (default one). I guess the only way is to write one by myself (which is actually not a problem at all), I am just wondering, maybe I missed something?

VladPodilnyk:
Also, it I've found that ` collection-compat ` is global lib, so it has been already added to a subproject.
@neko-kai

fundamentals/fundamentals-bio/src/test/scala/izumi/fundamentals/bio/test/SchedulerTest.scala
neko-kai: @VladPodilnyk You may use `Scheduler.global`, also monix's Timer is available implicitly `implicitly[Timer[monix.bio.Task]]` (You may find other components in MonixBIODefaultModule)

neko-kai:
Could we snatch some more test cases from e.g. https://github.com/zio/zio/blob/master/core-tests/shared/src/test/scala/zio/ScheduleSpec.scala ?

VladPodilnyk:
> Could we snatch some more test cases...

@neko-kai, I'll try.

fundamentals/fundamentals-bio/src/test/scala/izumi/fundamentals/bio/test/SchedulerTest.scala
VladPodilnyk: @neko-kai, thanks, I've summoned it implicitly. For some reason I thought compiler wouldn't find it...

fundamentals/fundamentals-bio/src/main/scala/izumi/functional/bio/retry/RetryPolicy.scala
```diff
@@ -0,0 +1,207 @@
...
+            ControllerDecision.Repeat(l -> r, interval, combined(next, RetryFunction.done(r)))
+          case (ControllerDecision.Stop(l), ControllerDecision.Repeat(r, interval, next)) =>
+            ControllerDecision.Repeat(l -> r, interval, combined(RetryFunction.done(l), next))
+          case (ControllerDecision.Stop(l), ControllerDecision.Stop(r)) => ControllerDecision.Stop(l -> r)
+        }
+    }
+    RetryPolicy(combined(action, policy.action))
+  }
+
+  def >>>[B1](that: RetryPolicy[F, B, B1]): RetryPolicy[F, A, B1] = {
```
neko-kai: @VladPodilnyk Could you check if you could move `: Monad2`/`: Applicative2` constraints down to method level instead of the class header?

```suggestion
  def >>>[B1](that: RetryPolicy[F, B, B1])(implicit F: Monad2[F]): RetryPolicy[F, A, B1] = {
```

VladPodilnyk: Yup, it is possible. 
Done ✅

neko-kai:
@VladPodilnyk Thank you! 🙏

PR #1612, opened 2021-10-13
Add "hideKeys" option

zawodskoj:
This PR allows displaying messages without key prefix using standard log formatter


neko-kai:
@zawodskoj Thank you for your contribution!! 🙏

PR #1615, opened 2021-10-14
Just-in-time non-sequential plan interpreter with IC prioritization

pshirshov:
This is a major cleanup which makes execution straightforward and removes a lot of unnecessary graph computations.

There are two major differences in the interpreter 

1. It's made aware of integration checks
2. It does not apply toposort anymore and traverses the graph directly, prioritizing integration checks

Still needs to be done:

- [x] Fix compile-time tests (not sure how did I break them, maybe that's unrelated to this PR) **???**
- [x] Integration checks are not prioritized globally, this can be improved (we may compute IC graph first when always prioritize IC paths while traversing). **EASY**
- [ ] Remove throwables from planner/interpreter **FAIR**
- [ ] Remove outdated code in test runner (I did basic cleanups but I'm sure more can be removed now)  **FAIR**
- [ ] Think how may we extract integration checks into separate/abstract interpreter hook/plugin/module **FAIR**

Should be a subject of a separate PR:

- [ ] ~~True parallel interpreter **HARD**~~ 
- [ ] ~~Efficient** integration check sharing/memoization? **HARD**~~


pshirshov:
Re broken compile-time test. In fact there is just one test which crashes the compiler:

```scala
  "Check with invalid role produces error" in {
    val result = PlanCheck.runtime.checkApp(StaticTestMain, PlanCheckConfig("unknownrole"))
    assert(result.maybeErrorMessage.exists(_.contains("Unknown roles:")))
    assert(result.issues.fromNonEmptySet.isEmpty)

    val err = intercept[TestFailedException](assertCompiles("""
      PlanCheck.assertAppCompileTime(StaticTestMain, PlanCheckConfig("unknownrole"))
      """.stripMargin))
    assert(err.getMessage.contains("Unknown roles:"))
  }

```

PR #1682, opened 2022-01-06
noticket: user specific container tags

VladPodilnyk:


distage/distage-framework-docker/src/main/scala/izumi/distage/docker/Docker.scala
```diff
@@ -194,13 +198,16 @@ object Docker {
...
+    * @param searchTag    A tag that is used to search containter. Useful when several containers are available to reuse
     */
   final case class ClientConfig(
     globalReuse: DockerReusePolicy = ClientConfig.defaultReusePolicy,
     useRemote: Boolean = false,
     useRegistry: Boolean = false,
     remote: Option[RemoteDockerConfig] = None,
     registry: Option[DockerRegistryConfig] = None,
+    searchTag: Option[String] = None,
   )
```
VladPodilnyk: Don't like this, but I actually haven't find a better way to pass a client search tag yet :/

distage/distage-framework-docker/src/test/scala/izumi/distage/testkit/docker/DistageTestDockerBIO.scala
```diff
@@ -71,3 +71,15 @@ final class DistageTestDockerBIOSecondEnv extends DistageTestDockerBIO {
 final class DistageTestDockerBIOThirdEnv extends DistageTestDockerBIO {
   override protected def config: TestConfig = super.config.copy(logLevel = Log.Level.Error)
 }
+
+// Labels tests
+final class DistageTestDockerTagsTest1 extends DistageTestDockerBIO {
```
neko-kai: @VladPodilnyk 
These tests only test launching the container, they're not testing if the tags are there - you may probably want to test that.

VladPodilnyk: Yup, thx!
I intended to do so, but I've got an issue with those test and I've hopped in to find a cause hence it's not fully complete.

VladPodilnyk: I've fixed tests, so they check if tag exist, but it doesn't check a situation where two containers with same image are available, but DIstage choose that one with tags

neko-kai (approved):

PR #1706, opened 2022-02-17
Cats Effect 3 Support

neko-kai:


IIVat:
@neko-kai thanks for your amazing efforts! Do you have any approximate estimation when the changes could be released?

neko-kai:
@IIVat 
Targeting this week.

IIVat:
Wow, that's awesome! I would like to bring distage into my project, but without CE3 this is a bit tricky. Thanks for the fast reply and  so encouraging information.

PR #1766, opened 2022-07-27
Experiment: convert any CE effect type to a bifunctor effect

neko-kai:
Seems to be working: cats-effect laws pass, but there may still be problems they don't reveal


mucaho:
Would it be possible to generalize this even further?

Namely, given any monofunctor `F[_]` and a corresponding cats type class instance e.g. `cats.Applicative[F]`, generate the respective BIO type class instance `[+E, +A] => izumi.functional.bio.Applicative2[Bifunctorized[F, E, A]]` for it.

Repeat that for all conversions in `izumi.functional.bio.CatsConversions`, just the other way around

neko-kai:
@mucaho
Sure, just need to make sure it's sound first.

mucaho:
Great to see this is still being worked on, thanks!

While this was being worked on, I have created my own two conversions I needed in the meantime, not sure if they are completely sound though, in case anyone needs: https://bitbucket.org/mucaho/ceffbanx/src/8986109fb7e9e97c7ae5a71ba311ff9d915634b8/src/main/scala/org/bitbucket/mucaho/ceffbanx/Util.scala#lines-130:180

neko-kai:
cats-mtl added "submarine error handling" in similar vein - https://github.com/typelevel/cats-mtl/pull/619 [associated blogpost](https://github.com/typelevel/typelevel.github.com/blob/4935d015fa2dfbf4867fb80dd90bcf733fb0dbae/collections/_posts/2025-08-17-custom-error-types.md)
The distinction between cats.Submarine and bio.TypedError is that Submarine uses an additional AnyRef to ensure valid nesting of handlers. Not sure whether that's a relevant addition for bifunctors.

neko-kai:
NB: nesting of handlers may not be relevant for bifunctorization. But distinguishing between different bifunctorized effect types may be. (we could just use TagK instead of path-dependency?)

PR #1837, opened 2022-12-09
tests can read activations from config

pshirshov:


Caparow (approved):

distage/distage-testkit-core/src/main/scala/izumi/distage/testkit/services/dstest/DistageTestRunner.scala
```diff
@@ -134,10 +135,34 @@ class DistageTestRunner[F[_]: TagK: DefaultModule](
   ): PackedEnv[F] = {
     // make a config loader for current env with logger
     val config = loadConfig(env, configLoadLogger)
+
     val lateLogger = EarlyLoggers.makeLateLogger(RawAppArgs.empty, configLoadLogger, config, envExec.logLevel, defaultLogFormatJson = false)
 
+    val fullActivation = env.activationStrategy match {
```
neko-kai, line 141: @pshirshov 
Eh, this means tests are sorted into memoization envs without taking config into account and then config is added, which can lead to wrong components being memoized into envs

pshirshov: That shouldn't be the case, TestConfig should be a part of the memoization key and the config should be the same for every classpath.

PR #1859, opened 2023-01-12
Cross-test logger memoization

pshirshov:


pshirshov:
I guess it's lot better now. The implementation is cleaner and does not involve a global singleton but there is no logger sharing across effects (which shouldn't affect 99% of the real user code).

Caparow (approved):

PR #1861, opened 2023-01-16
Distage: Testkit memoization tree cleanup

Caparow:
@neko-kai I think we can remove bootstrap module rewrite here:
```scala
    val locatorWithOverriddenPlannerAndActivationInfo: LocatorDef = new LocatorDef {
      // we override ActivationInfo <s>(& Activation)</s> because the test can have _different_ activation from the memoized part
      // FIXME: Activation will be part of PlannerInput in 0.11.0 & perhaps ActivationInfo should be derived from Bootloader/PlannerInput as well instead of injected externally      
```

### DIFF
--- a/distage/distage-core-api/src/main/scala/izumi/distage/model/definition/errors/DIError.scala
+++ b/distage/distage-core-api/src/main/scala/izumi/distage/model/definition/errors/DIError.scala
@@ -19,7 +19,7 @@ sealed trait DIError
 
 object DIError {
   implicit class DIResultExt[A](private val result: Either[List[DIError], A]) extends AnyVal {
-    def cumulateErrors: Either[InjectorFailed, A] = {
+    def aggregateErrors: Either[InjectorFailed, A] = {
       result match {
         case Left(errors) =>
           val i = new DIFailureInterpreter()
@@ -29,7 +29,7 @@ object DIError {
           Right(resolved)
       }
     }
-    def getOrThrow(): A = cumulateErrors match {
+    def getOrThrow(): A = aggregateErrors match {
       case Left(value) => throw value
       case Right(value) => value
     }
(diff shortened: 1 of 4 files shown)


distage/distage-core-api/src/main/scala/izumi/distage/model/definition/errors/DIFailureInterpreter.scala
```diff
@@ -23,21 +27,21 @@ class DIFailureInterpreter() {
 
     val inconsistencies = issues.collect { case e: DIError.VerificationError => DIError.formatError(e) }.niceList()
     if (inconsistencies.nonEmpty) {
```
neko-kai: missing `else`, all rewritten throws above are ignored right now.

Caparow: 🤦

distage/distage-core-api/src/main/scala/izumi/distage/model/definition/errors/DIError.scala
```diff
@@ -18,16 +19,20 @@ sealed trait DIError
 
 object DIError {
   implicit class DIResultExt[A](private val result: Either[List[DIError], A]) extends AnyVal {
-    def getOrThrow(): A = {
+    def cumulateErrors: Either[InjectorFailed, A] = {
```
pshirshov: aggregate or accumulate I think

pshirshov (approved):

distage/distage-core-api/src/main/scala/izumi/distage/model/definition/errors/DIError.scala
Caparow: 👍

PR #1989, opened 2023-08-03
Remove ZManaged

felix-hedenstrom:
Distage currently has a required dependency on `ZManaged` if you want to use ZIO 2 with `Lifecycle`. This PR removes the `ZManaged` dependency.

This is change is done because ZIO no longer recommends using `ZManaged` and instead prefers `ZIO[Scope, _, _]`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3036,7 +3036,6 @@ lazy val `distage-framework-docker` = project.in(file("distage/distage-framework
       "org.typelevel" %% "cats-core" % V.cats % Test,
       "org.typelevel" %% "cats-effect" % V.cats_effect % Test,
       "dev.zio" %% "zio" % V.zio % Test excludeAll("dev.zio" %% "izumi-reflect"),
-      "dev.zio" %% "zio-managed" % V.zio % Test excludeAll("dev.zio" %% "izumi-reflect"),
       "dev.zio" %% "izumi-reflect" % V.izumi_reflect % Test,
       "com.github.docker-java" % "docker-java-core" % V.docker_java,
       "com.github.docker-java" % "docker-java-transport-zerodep" % V.docker_java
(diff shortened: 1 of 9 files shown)


neko-kai:
@felix-hedenstrom 
Hi, thank you very much for your contribution! 🙏  I am not sure about fully removing support for ZManaged (instead of leaving at as Optional) - even though libraries have largely moved to Scope, after some deliberation I think I'd still prefer to have an option to use it with distage, because I personally prefer it over Scope, if I were to use ZIO directly without BIO typeclasses.

Are you interested in pursuing this PR to completion? There are some build failures left and if you do want to continue I can guide you on how to make ZManaged fully Optional. Otherwise I can take this from here.

felix-hedenstrom:
Thanks for the reply. I can respect your preference of ZManaged. My company has tried to move away from it and it was a pain point to require our ZIO[Scope or ZLayer to be transformed into ZManaged if we wanted to continue using distage.

I got a bit stuck trying to solve the remaining build issues and sadly have used up the allotted time for this PR in the foreseeable future. I think it would be useful if you continued the PR from here.

neko-kai:
@felix-hedenstrom 
Note that ZLayer is already supported by `.fromZEnv` method - there's no need to convert ZLayer to ZManaged. For scoped ZIO you can use the following extension methods for now as a workaround, until `.fromResource` and `.fromZEnv` support Scope natively:

```scala
    implicit class MakeFromZIOScopeResource[T, AfterBind](protected val dsl: MakeDSLBase[T, AfterBind]) extends AnyVal {

      def fromZIOScopeResource[E: Tag, I <: T: Tag](resource: ZIO[Scope, E, I]): AfterBind = {
        dsl.fromResource(ZManaged.scoped(resource))
      }

      def fromZIOScopeResource[E: Tag, I <: T: Tag](function: Functoid[ZIO[Scope, E, I]])(implicit d1: DummyImplicit): AfterBind = {
        dsl.fromResource(function.map(ZManaged.scoped(_)))
      }

    }
```

Okay, I'll continue from here. Thank you for the effort!

neko-kai:
Fixed in https://github.com/7mind/izumi/pull/2010

PR #2058, opened 2024-01-29
Bug: ConfigWriter minimizer does not omit nested unrequired nodes.

Caparow:


pshirshov:
This is not a bug, it's a fundamental limitation of our codec model. Current codecs cannot provide metadata w/o execution and even during execution they cannot trace all possible code paths because the structures are polymorphic. So, we can't solve this perfectly w/o complete rethinking of our codec model (which is the same as in circe/pureconfig).

PR #2082, opened 2024-03-07
Update to Scala 3.4.0, memoization fixed on Scala 3 with update

neko-kai:


SethTisue:
It came up today on Discord that someone was trying to use Izumi 1.2.6 on Scala 3.3, and that didn't work because of the 3.4 upgrade here. I'm curious if the 3.4 upgrade was somehow actually necessary, or if you weren't aware that Scala 3 libraries normally choose to continue supporting the last LTS release which was 3.3.

pshirshov:
From what I can remember, there is a critical bug in 3.3 which breaks identity of izumi-reflect tags. We would be happy to stay on 3.3 but we can't.

neko-kai:
@SethTisue We depend on the fix to https://github.com/scala/scala3/issues/19224 which was released in 3.4.0. It's not quite viable for us to support older versions due to this. I hope they don't critically need 3.3 LTS support, and can update after all. They could permanently stay on an older version (and experience bugs in testkit due to above issue), but that would also be a very bad idea.

We are aware that it's probably best not to update Scala 3 when possible, our other libraries (izumi-reflect), still publish using 3.2

pshirshov:
We can downgrade, if the patch is backported into LTS.

SethTisue:
I see. Maybe worth a mention in readme and/or doc page? Not sure.

> if the patch is backported into LTS.

Looks to me like the fix is included in 3.3.4-RC1.

neko-kai:
@SethTisue Yes, we've been able to downgrade to 3.3.4-RC1 just now. Thanks for the tip!

PR #2106, opened 2024-04-22
Functoid isolation

pshirshov:


distage/distage-core-api/src/main/scala-2/izumi/distage/model/definition/dsl/LifecycleTagMacro.scala
```diff
@@ -8,7 +8,7 @@ import scala.reflect.macros.blackbox
 object LifecycleTagMacro {
   def fakeResourceTagMacroIntellijWorkaroundImpl[R <: Lifecycle[Any, Any]: c.WeakTypeTag](c: blackbox.Context): c.Expr[Nothing] = {
     val tagMacro = new TagMacro(c)
-    tagMacro.makeTagImpl[R] // run the macro AGAIN, to get a fresh error message
+    val _ = tagMacro.makeWeakTag[R] // run the macro AGAIN, to get a fresh error message
```
neko-kai: I think this should remain `makeTagImpl`, otherwise there would be no update to error message in `tagMacro.getImplicitError`

PR #2120, opened 2024-05-24
Use scalatest Runner.discoveredSuites, remove all custom discovery

### LINKED ISSUE #1055: distage-testkit-scalatest: to work around `sbt testOnly` launching all the tests, add `-- -z ""` parameter at the end
# Upstream fix

Upstream was fixed long ago, but still not published: https://github.com/scalatest/scalatest/issues/2375

# Current workaround

> I've found that you may recover testOnly with the default implementation by adding  -- -z "" boilerplate at the end! e.g. sbt> testOnly abc.MyTest -- -z "" - would behave just as expected, running only MyTest.

# Problem 

`sbt testOnly` doesn't work correctly with distage-testkit-scalatest – ALL tests run, instead of just the selected.

# Workaround

Instead of regular invocation

```
sbt> testOnly org.mypackage.MyTestClass
```

Add options `-- -z ""` at the end of `testOnly` command as follows:

```
sbt> testOnly org.mypackage.MyTestClass -- -z ""
```

Then `testOnly` works as expected

# Details

As detailed in https://gitter.im/7mind/izumi?at=5ebbf553af599706dff55531 `sbt testOnly` doesn't work for distage-testkit-scalatest tests – because Scalatest does not provide data to suites about which _other_ suites it's going to run – and in sbt-mode there's no way to figure that out via side-channels, like distage-testkit does in Intellij IDEA / vanilla ScalatestRunner (single suite launch works there) there's no choice but to run _all_ the suites.

We can allow the user to workaround that by applying the pattern supplied in `-z` or `-t` to test Suites also - instead of `testOnly AbcTest` write `testOnly -- -n AbcTest`. Additionally it may be possible (but is unlikely) to figure out if we're running in `testOnly` and throw an error in that case, advising a workaround.

related issue: we should open an issue to notify upstream and/or have an issue to link to.

neko-kai:
https://github.com/scalatest/scalatest/pull/2319/

Fix #1055

Updates to scalatest 3.3.0-alpha.2. This will require you to update scalatest extensions and org.scalatestplus modules if you use any to 3.3.0-alpha.2 / 3.3.0.0-alpha.2


neko-kai:
Still waiting for upstream to release a new version 💀 https://github.com/scalatest/scalatest/issues/2375

PR #2221, opened 2024-11-29
Relax restriction on .modify call after .from in ModuleDef DSL, allow widening implementation type. Apply .namedByImpl after modifications, not before

neko-kai:


neko-kai:
Hmm, has implications on losing integration checks after .modify. Maybe after https://github.com/7mind/izumi/issues/1658 ?

Hmm, can't ordinary .modify also lose integration checks? (it can't after https://github.com/7mind/izumi/pull/2255)

PR #2225, opened 2024-12-11
Subcontext: add `unsafeModify`. Open `SubcontextImpl` class. Rename SubcontextImpl `.empty`->`.initial`

neko-kai:


neko-kai:
Actually we can't really add `flatAp`, because we can't expand the set of roots - the plan is already pre-planned with current roots. Any new components requested in flatAp would have to have already been dependencies of `A` or they wouldn't be found.

PR #2232, opened 2024-12-23
extractSelf convenience helper

pshirshov:
I'm not sure if this is the best option. Maybe we should do this by default and override with `extractWith`.


pshirshov:
Ah, it's already like this since last refactoring. Maybe we still should add this method to allow users to have more clarity in their code.

PR #2264, opened 2025-03-19
DSL: Added .named methods with optional identifier

YaroslavSagach:


distage/distage-core/src/test/scala/izumi/distage/injector/BasicTest.scala
```diff
@@ -595,4 +595,34 @@ class BasicTest extends AnyWordSpec with MkInjector with ScalatestGuards {
...
+  "support named bindings with option" in {
+    import BasicCase10.*
+    import SetCase4.*
+    val definition = PlannerInput.everything(new ModuleDef {
+      make[TestClass]
+      make[TestGreeter].named(Some("named.greeter"): Option[Identifier])
+      make[TestGreeter].named(None) // should bind without id
+      make[TestDependency].named(Some("named.test.before.from"): Option[Identifier]).from[TestImpl1]
+      make[TestDependency].named(None).from[TestImpl1]
+      make[TestDependency].from[TestImpl2].named(Some("named.test.after.from"): Option[Identifier])
```
neko-kai: @YaroslavSagach Try to change signature to `def named[I: IdContract](name: Option[I])` - maybe this way you may be able to avoid writing `: Option[Identifier]`

neko-kai (changes requested):

distage/distage-core/src/test/scala/izumi/distage/injector/BasicTest.scala
YaroslavSagach: In that case we need to write a type, in case of None - `make[TestDependency].named[String](None)`, because of ambiguity in IdContract implicits.
If that is better, I will leave the implementation with IdContract

neko-kai: @YaroslavSagach You can fix that by refactoring IdContract implicits like this:

```scala

object IdContract extends LowPriorityIdContractInstances {
  ...

  implicit def singletonStringIdContract[S <: String & Singleton]: IdContract[S] = new IdContractImpl[S]
}
sealed trait LowPriorityIdContractInstances {
  implicit final val stringIdContract: IdContract[String] = new IdContractImpl[String]
}
```
Then `singletonStringIdContract` gets chosen by default.

neko-kai (approved):

PR #2268, opened 2025-04-17
Distage: Added test for trait generation with constructors + sort intersection members

YaroslavSagach:


fundamentals/fundamentals-platform/src/main/scala-3/izumi/fundamentals/platform/reflection/ReflectionUtil.scala
```diff
@@ -45,7 +45,7 @@ object ReflectionUtil {
     }
 
     go(typeRepr)
-    tpes.toList
+    tpes.toList.sortBy(_.typeSymbol.flags.is(Flags.Trait))
```
neko-kai: I think better move the sort [here](https://github.com/7mind/izumi/blob/702c9320c617bc7eeacf93dabc838e88c80b354e/distage/distage-core-api/src/main/scala-3/izumi/distage/constructors/ConstructorUtil.scala#L23) and add a comment for why it's necessary. The ReflectionUtil.intersectionMembers function is also used in HasConstructor and is a bit more general, it's somewhat weird for it to reorder intersection arguments on its own.

neko-kai (approved):

PR #2283, opened 2025-06-05
Bind functions with implicit parameters

nikita-ty754:


distage/distage-testkit-scalatest/src/main/scala/izumi/distage/testkit/scalatest/Spec1.scala
```diff
@@ -8,7 +8,7 @@ import org.scalatest.distage.DistageScalatestTestSuiteRunner
 
 import scala.language.implicitConversions
 
-abstract class Spec1[F[_]: TagK: DefaultModule] extends DistageScalatestTestSuiteRunner[F] with DistageAbstractScalatestSpec[F] {
+abstract class Spec1[F[_]](implicit tag: TagK[F], module: DefaultModule[F]) extends DistageScalatestTestSuiteRunner[F] with DistageAbstractScalatestSpec[F] {
```
ntyshchenko-playq: This change is required to make MyDisabledTestF compile on both scala 3.7.0 and 2.13.16
```
abstract class MyDisabledTestF[F0[_]: QuasiIO: DefaultModule, F[x] <: F0[x]: TagK](f0Tag: TagK[F0])(implicit F: Applicative[F]) extends Spec1[F0]()(f0Tag, implicitly[DefaultModule[F0]]) {
```
Maybe it would be better to have separate tests for scala 3 and 2 instead of changing Spec1 signature ? In case we are okey with this change than it makes sense to change signatures for Spec2 and Spec3

neko-kai:
I'll add a documentation chapter to basics.md <strike>before</strike> after merging

PR #2331, opened 2025-12-22
Scala 3: Avoid re-typing of macro results in CodePosition macros by wrapping everything in `Typed` nodes

neko-kai:
Confirmed in https://github.com/7mind/izumi/pull/2325 that using Typed nodes avoids typecheck step on macro result - same as `setType` on Scala 2

### DIFF
--- a/fundamentals/fundamentals-language/src/main/scala-3/izumi/fundamentals/platform/language/CodePositionMaterializer.scala
+++ b/fundamentals/fundamentals-language/src/main/scala-3/izumi/fundamentals/platform/language/CodePositionMaterializer.scala
@@ -54,6 +54,7 @@ object CodePositionMaterializer {
         .map(_.trim)
         .mkString(".")
 
+      // Use Typed nodes to avoid retypechecking (this required internal.setType on Scala 2, but seems like Typed(..) achieves the same on Scala 3)
       Typed(Literal(StringConstant(applicationId)), TypeTree.of[String]).asExpr.asInstanceOf[Expr[String]]
     }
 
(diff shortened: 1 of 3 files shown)


neko-kai:
I'm misunderstanding something. Dotty macro trees appear to all have types set at construction already, so are we gaining any performance whatsoever by using Typed nodes OR setting type with `dotty.tools.dotc.ast.Trees.Tree.withType`? Seems like no, after `withType` we're getting the exact same instance of a tree as before (i.e. no copying happened). The only difference between using a Typed node and `withType` is that Typed will insert a checked cast, while `withType` will allow incorrect bytecode to be output and cause a `java.lang.VerifyError` at runtime.

neko-kai:
No action seems to be necessary, all tree node types are already set, checking with:
```scala
def checkAllTyped(using qctx: Quotes)(t: qctx.reflect.Term): Unit = {
  import qctx.reflect.*
  val visited = mutable.ListBuffer.empty[Any]
  val isTyped = new TreeAccumulator[Boolean] {
    override def foldTree(x: Boolean, tree: qctx.reflect.Tree)(owner: qctx.reflect.Symbol): Boolean = {
      val bool = x && tree.asInstanceOf[untpd.Tree].hasType && tree.asInstanceOf[untpd.Tree].tpe.ne(null)
      if (bool) {
        visited += tree
        foldOverTree(bool, tree)(owner)
      } else {
        println(s"untyped $tree")
        false
      }
    }
  }.foldTree(true, t)(Symbol.noSymbol)
  if (isTyped) {
    println(s"all typed: ${visited.zipWithIndex.map((s, i) => s"${(" " * i)}- ${s.toString}").mkString("\n")}")
  } else {
    println(s"untyped in $t")
    throw new RuntimeException("untyped")
  }
}
def checkAllTyped[A](using qctx: Quotes)(e: Expr[A]): Expr[A] = {
  import qctx.reflect.*
  checkAllTyped(e.asTerm)
  e
}
```

PR #2346, opened 2026-04-09
distage-framework-docker: Add platform to pull command, will fix #2345

### LINKED ISSUE #2345: distage-framework-docker: distage-docker: propagate platform to image pull when autoPull is enabled
## Problem

`ContainerResource` applies `config.platform` when creating a container, but does not apply the same platform when pulling the image.

Current behavior in `ContainerResource`:

- `createContainerCmd(...).withPlatform(...)` is used when `config.platform` is defined
- `pullImageCmd(...)` does not receive `config.platform`

This can cause inconsistent behavior on multi-architecture hosts.

For example, on Apple Silicon, if `autoPull = true` and `platform = Some("linux/amd64")`:

1. the image pull may default to the host platform (`linux/arm64`)
2. the container create step then requests `linux/amd64`

That means pull and create are using different platform selection rules.

## Why this looks like a bug

`platform` is already part of `Docker.ContainerConfig`, so from the API perspective it looks like a container-level platform preference that should apply consistently across the full startup flow, not only at container creation time.

Also, the underlying dependency (`docker-java 3.6.0`) already supports `PullImageCmd.withPlatform(...)`, so this is not blocked by upstream API limitations.

A quick look at git history suggests this was likely an omission rather than an intentional design choice: the commit that introduced `platform` support added it to the config and to `createContainerCmd`, but not to `pullImageCmd`.

## Proposed fix

Propagate `config.platform` into the pull step as well, so the image is pulled for the same platform that is later requested during container creation.

Conceptually:

```scala
val pullCmd = Value(rawClient.pullImageCmd(imageName))
  .mut(config.platform)(_.withPlatform(_))
  .mut(registry)(_.withRegistry(_))
  .mut(registryAuth)(_.withAuthConfig(_))
  .get
```

## Expected behavior after the fix
When platform is set, both:

image pull
container creation
should use the same platform value.

When platform is not set, behavior should remain unchanged.

ETCHKILI:
Fix #2345


neko-kai:
@ETCHKILI Merged, thanks! :+1: Although I replaced your reflective test with an end-to-end one

neko-kai:
@ETCHKILI Note: your changes are published as part of 1.3.0-M1 release if you need them live

PR #2349, opened 2026-04-25
Deprecate `make` without `.from`, use `make[T].fromSelf` instead

neko-kai:
Supporting `make[T]` syntax without a follow-up `from` is rife with issues (see https://github.com/scala/scala3/issues/21622) Ports of distage to [Python](https://github.com/7mind/izumi-chibi-py) and [TypeScript](https://github.com/7mind/izumi-chibi-ts) do not support solitary `make` and that seems to work out just fine.

Commit is generated and needs cleanup/review before merge.


neko-kai:
Maybe we should use `makeClass[T]` instead of, or in addition to, `.fromSelf`. We already have `makeTrait[T]`

